PR #1, opened 2024-03-20
MSEOSKERN-4482: Profiling Agent Command Line Arguments

gmarler:
Expand or improve our current Clap 4 Derive command line arguments:

* Specify TIDs to profile - in addition to the PIDs we select
  * These can be outside/beyond the PIDs that have been selected, if any
* Accept per-CPU raw stack sampling frequency in Hz
   * Defaults to 19 Hz
   * Limit to range 1 Hz to 1009 Hz
   * Ensure selected number is a prime number
* Duration of the profiling agent run in seconds
   * Defaults to forever, or 'continuous'
    * Thus, we eliminate the continuous option
 * The name of the file in which the emitted flamegraph SVG will be placed
     * Defaults to "flame.svg"
* Need to add an option that expresses at which interval or rate "collections" occur


src/main.rs
```diff
@@ -12,27 +13,72 @@ use lightswitch::object::build_id;
...
+    Ok(Duration::from_secs(seconds))
+}
+
+const SAMPLE_FREQ_RANGE: RangeInclusive<usize> = 1..=1001;
+
+fn sample_freq_in_range(s: &str) -> Result<u16, String> {
+    let sample_freq: usize = s
+        .parse()
+        .map_err(|_| format!("`{s}' isn't a valid frequency"))?;
+    if !(1..=1001).contains(&sample_freq) {
```
javierhonduco, line 33: I assume you wanted to use the constant here?
```suggestion
    if !SAMPLE_FREQ_RANGE.contains(&sample_freq) {
```

src/main.rs
```diff
@@ -12,27 +13,72 @@ use lightswitch::object::build_id;
...
+        return Err(format!(
+            "sample frequency not in allowed range {}-{}",
+            SAMPLE_FREQ_RANGE.start(),
+            SAMPLE_FREQ_RANGE.end()
+        ));
+    }
+    if !is_prime(sample_freq.try_into().unwrap()) {
```
javierhonduco, line 40: What do you think about giving a warning `warn!("msg")` rather than preventing the user from using a nonprime? No strong opinions here, though

src/main.rs
```diff
@@ -12,27 +13,72 @@ use lightswitch::object::build_id;
...
+        return Err(format!("sample frequency is not prime"));
+    }
+    Ok(sample_freq as u16)
+}
+
+// Used to prevent the default value for duration from being dropped before
+// we parse the command line args - probably a better way to do this
+fn max_duration_as_str() -> &'static str {
+    let max_duration: &'static str =
+        Box::leak(Duration::MAX.as_secs().to_string().into_boxed_str());
```
javierhonduco, line 50: I am not sure if the `Box::leak` is needed here... maybe this could work?

```
fn max_duration_as_str() -> String {
    Duration::MAX.as_secs().to_string()
}
```

src/profiler.rs
```diff
@@ -558,7 +563,7 @@ impl Profiler<'_> {
                             }
                         } */
                     } else {
-                        error!("unknow event {}", event.type_);
```
javierhonduco, line 561: oops!

javierhonduco (commented):
LGTM! Just left some minor comments. Let's try to remove the `Box::leak` though 😄

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 flame.svg
-src/bpf/*_skel.rs
\ No newline at end of file
+src/bpf/*_skel.rs
+*.lock
```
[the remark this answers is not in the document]
gmarler: Unrelated - I was going to ask why we're keeping the *.lock files in git.

src/main.rs
gmarler: The important thing is that the final result is a Str/str, NOT a String - clap requires string-like default values to always be some form of str/Str.

src/main.rs
```diff
@@ -12,27 +13,72 @@ use lightswitch::object::build_id;
...
 use std::path::PathBuf;
 
 use std::time::Duration;
 
+fn parse_duration(arg: &str) -> Result<Duration, std::num::ParseIntError> {
+    let seconds = arg.parse()?;
+    Ok(Duration::from_secs(seconds))
+}
+
+const SAMPLE_FREQ_RANGE: RangeInclusive<usize> = 1..=1001;
```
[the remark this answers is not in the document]
gmarler: Yup, fixed

src/main.rs
gmarler: Ugh - yes, fixed.

src/main.rs
gmarler: That would be a fine expansion - I also had the idea to show the nearest primes above and below a non-prime selection attempt.  A future PR/bug fix I suppose.

.gitignore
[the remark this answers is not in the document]
javierhonduco: The `Cargo.lock` and the `flakes.lock` files ensure that the dependencies that we use both for Rust and Nix are the same across invocations and systems

src/main.rs
javierhonduco: Can we convert it into that type in the clap code? So call `to_string()` in the macro?

gmarler: Finally discovered how using the `string` feature for the clap crate makes this work for you.  You can now use a String, and it'll auto-persist it and provide a pointer to the str/Str slice within it.  So this is all fixed, and `max_duration_as_str()` is no longer necessary and has been removed.

javierhonduco: Amazing! Thanks so much for improving this 🙌

PR #5, opened 2024-03-25
Fixes for 20240325

gmarler:
This fixes 2 separate issues:

1. The eBPF verifier in newer kernels requires all items on the stack to be fully initialized.  This generally requires all values on the stack to be 64 bit values, or a multiple of 64 bits.  Using u64 variables rather than u32 guarantees we won't have issues at present.
2. The variable name PERF_BUFFER_PAGES stores a value that is actually in units of bytes, so changing the name to PERF_BUFFER_BYTES


javierhonduco (approved):
Great stuff, thanks for verifying that the `zero` being 32 bits was the issue on your Ubuntu box!

PR #9, opened 2024-04-06
Give helpful hint if sampling frequency isn't prime

gmarler:
Currently, we exit with error if the `--sample-freq` option is provided with a non-prime number.

This prevents accidentally sampling at rates that could collide with other system events more easily, leading to unintended feedback.

But, it doesn't help the user know what numbers they should have picked.

This PR adjusts the error message to include the prime numbers below and above the sampling rate that was requested.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,9 +38,39 @@ fn sample_freq_in_range(s: &str) -> Result<u16, String> {
         ));
     }
     if !is_prime(sample_freq.try_into().unwrap()) {
-        return Err("sample frequency is not prime".to_string());
+        // return Err("sample frequency is not prime".to_string());
+        let ba_result = primes_before_after(sample_freq.try_into().unwrap());
+        match ba_result {
+            Ok((prime_before, prime_after)) => {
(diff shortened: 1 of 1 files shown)


javierhonduco:
This looks good but do you mind separating it in one PR per feature, perhaps one for the primes, one for the new argument, etc?

src/main.rs
```diff
@@ -105,6 +105,9 @@ struct Cli {
     /// Output file for Flame Graph in SVG format
     #[arg(long, default_value = "flame.svg")]
     flamegraph_file: PathBuf,
+    /// Enable eBPF debug output
+    #[arg(long, action=ArgAction::SetFalse)]
```
javierhonduco: nit: this is not the BPF logs but the `libbpf` ones

gmarler:
> This looks good but do you mind separating it in one PR per feature, perhaps one for the primes, one for the new argument, etc?

Sure thing - I've reverted that change in this branch - we probably need to discuss it anyway, as in BPF vs libbpf logs.

PR #14, opened 2024-04-15
Mappings redesign

javierhonduco:
## Redesign mapping storage and search

Rather than keeping a large-ish fixed-size array for every single
process, use the longest prefix match that the BPF trie data structure
provides, this has several benefits.

- Searching is no longer linear the number of mappings;
- The max number of mappings is lifted, and now the total amount of
  mapping entries are shared among the processes, so processes that have
few mappings will have few entries, while processes with lots of
mappings can have more entries than the current maximum, allowing
lightswitch to profile more applications;

### Future work

There are no counters in BPF or userspace in case of errors. There are a
bunch of areas in the code that can panic, such as map operations that
will fail if the maps are full.

### Test Plan

Added unit and integration tests, as well as ran this commit for several
hours with many different processes without issues.

# Rework known process detection
We don't need a separate map anymore. As BPF tracing programs require
hash maps to be preallocated, this will save some memory.

Something we might want to do in the future is tune the maximum entries
in the mappings BPF map.

Additionally, this also cleans up some of the implementations for the
BPF bindings. I love bindgen!!

### Test Plan

Ran locally without any issues or errors.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,6 +16,7 @@ fn main() {
     println!("cargo:rerun-if-changed={PROFILER_BPF_SOURCE}");
 
     let bindings = bindgen::Builder::default()
+        .derive_default(true)
         .header(PROFILER_BPF_HEADER)
         .generate()
         .expect("Unable to generate bindings");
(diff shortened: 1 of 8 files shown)


javierhonduco:
cc @gmarler as this is a significant rework on the way memory mapping work in the BPF side, let me know if you have any feedback!

PR #16, opened 2024-04-19
Add support for reading headers of 32 bit ELF executables

javierhonduco:
As reported by Gordon Marler cc @gmarler 

Test Plan
=========

Ran locally without issues, will test in prod and see what the error messages are after this change.


javierhonduco:
Let me know if this fixes the issue you've seen, Gordon!

PR #19, opened 2024-04-23
symbolizer: Handle addresses with many symbols

javierhonduco:
Symbolization was completely broken if inline functions were present because the code assumed that every program counter maps to exactly one symbol, but that's not true.

We didn't see this often because some applications we ran had no inlined functions, but it was very apparent with Ruby and Fortran code, especially with O2 and O3, as the compiler will start inlining at these levels more.

Test Plan
=========

Tested with `basic_f77_stack_O2`.

**before**
image

**after**

image


javierhonduco:
cc @gmarler who reported this issue!

PR #25, opened 2024-04-24
Add kernel tests

javierhonduco:
This commit adds kernel tests running on x86_64 for 5.15, 6.0, 6.2, and 6.6 using vmtest[0]. The main goal is to ensure that the BPF programs load fine for various kernels.

The original plan was to have a nix derivation that would do everything: 1) download the kernels 2) build vmtest 3) build lightswitch 4) generate the vmtest configuration, and finally 5) run vmtest with said configuration. I attempted to do this in this branch [1] but I was taking too long and decided for this approach that gets us 80% of the value for 20% of the work.

The downsides for this approach are:

1) As the vmtest config file is not generated on the fly, the kernel
   paths have to be hardcoded. This is not as bad as it might seem
because nix guarantees these paths to be the same accross machines, but it's still a bit kludgy;
2) There is an expectation that lightswitch in dev mode has been built
   before runnign the vmtests;

How to run the kernel tests
===========================

```
nix develop
cargo build
vmtest
```

Notes
=====

We still require qemu-guest-agen to be running in the base OS (nix or not) which is not ideal, but a proper fix might be quite involved.

Test Plan
=========

Forced a panic to make sure the kernel tests would fail.

[0]: https://github.com/danobi/vmtest
[1]: https://github.com/javierhonduco/lightswitch/commits/wip-vmtests


javierhonduco:
cc @danobi 😊

javierhonduco:
Merging the changes, but if you have any feedback @gmarler, I am happy to change things. Let me know if you have any issues running the kernel tests on your box

PR #28, opened 2024-04-30
WIP: Performance improvements for unwind information

javierhonduco:
⚠️ Do not merge yet ⚠️ 

* modify the unwind info in place during redundant info removal and other optimizations (this showed a ~30% reduction in CPU time);
* sort out fdes rather than the whole unwind info (in benchmarks this showed no performance improvements, but in machines with less memory with swap enabled, it should help);

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,6 +121,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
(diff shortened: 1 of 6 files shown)


javierhonduco:
Superseded by https://github.com/javierhonduco/lightswitch/pull/30

PR #29, opened 2024-05-01
tests: Fix off-by-32 bits in LPM tree

javierhonduco:
The integration tests started failing starting with kernel 6.8, and after some checks, it pointed to this commit
https://github.com/torvalds/linux/commit/9b75dbeb36fcd9fc7ed51d370310d0518a387769.

The tests fail now because the prefix length we pass is 32 bits higher (see initialiser) than the maximum prefix length. After this commit was introduced, any prefix length that's higher than the key size will return NULL earlier, rather than returning the value with the longest match, even if it's beyond the size the key can encode.

Test Plan
=========

Ran the tests and some profiles on a box running 6.8.7-200 (Fedora) without issues.

Also added an assertion to ensure that the prefix_len is not bigger than the LPM trie key size.


javierhonduco:
Many thanks to @florianl for noting that this was not a kernel regression, but a bug on my code!

PR #31, opened 2024-05-02
Improve unwind debugging

gmarler:
This improves unwind debugging to a degree:
* implement Display trait for RawAggregatedSample
* Implement Display trait for SymbolizedAggregatedSample
* Since Linux PIDs are all actually threads, each folded stack should include both

    1. The process binary name (the `comm` field from `/proc/<PGID>/stat`)
    1.  The thread name, defaulting to **main_thread** for any given PGID - eventually, for cases where all threads have the same `comm`, we should name the threads as **<binary_name>_thread_X**, where X could be abs(PGID - PID).


src/profiler.rs
```diff
@@ -1110,6 +1167,14 @@ impl Profiler<'_> {
...
     pub fn setup_perf_events(&mut self) {
         let mut prog_fds = Vec::new();
+        // TODO: The call to num_possible_cpus is afflicted by https://github.com/libbpf/libbpf/issues/103
+        // Seeing examples of hosts where the number of possible CPUs is not the actual number of
+        // online CPUs:
+        // $ cat /sys/devices/system/cpu/possible
+        // 0-7
+        // $ cat /sys/devices/system/cpu/online
+        // 0-3
+        // So we need the number of online, not possible CPUs, to avoid failures
```
javierhonduco, line 1177: Thanks for documenting this! How did you notice this, I guess it panicked while trying to set the perf counter on that CPU?

Given that individual CPUs can be disabled perhaps we should check what are the actually online CPUs (as the range might be wrong). We could also do something a bit more scrappy and try to enable from `0..possible` and if there are errors continue to the next CPU. It should be fine as long as we can create the perf event in at least 1 CPU.

What do you think?

src/profiler.rs
```diff
@@ -535,12 +593,11 @@ impl Profiler<'_> {
                     }
 
                     let raw_sample = RawAggregatedSample {
-                        pid: key.task_id,
+                        pid: key.pid,
```
javierhonduco, line 596: Thanks for the fix! We should probably pass both the `pid` and `task_id` but we can do this in another PR if you prefer!

src/profiler.rs
```diff
@@ -142,6 +175,29 @@ pub struct SymbolizedAggregatedSample {
...
+                scratch_string.push_str(&cvtd);
+            }
+            scratch_string.push(']');
+            scratch_string
+        };
+        let ustack_rep = format_symbolized_stack(&self.ustack);
+        let kstack_rep = format_symbolized_stack(&self.kstack);
+        let final_rep = write!(
+            f,
+            "SymbolizedAggregatedSample:\npid: {}\nustack: {}\nkstack: {}\ncount: {}",
```
javierhonduco, line 194: This is lovely! Some small improvement we could make here is to use the formatting facilities from `fmt::Formatter`, something like:

```suggestion
             f.debug_struct("SymbolizedAggregatedSample")
                        .field("pid", &pid)
                        .field("ustack", &ustack)
                        .finish()
```

src/profiler.rs
```diff
@@ -126,13 +127,45 @@ const MAX_CHUNKS: usize = MAX_UNWIND_TABLE_CHUNKS as usize;
...
 }
 
+impl fmt::Display for RawAggregatedSample {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let format_native_stack = |native_stack: Option<native_stack_t>| -> String {
+            let mut scratch_string: String = String::new();
+            match native_stack {
+                None => scratch_string.push_str("NONE"),
+                Some(native_stack) => {
+                    scratch_string.push_str("[\n");
```
javierhonduco, line 145: nit: we could have this as a vector and then use `join("\n")` instead

src/profiler.rs
```diff
@@ -126,13 +127,45 @@ const MAX_CHUNKS: usize = MAX_UNWIND_TABLE_CHUNKS as usize;
...
+                    scratch_string.push(']');
+                }
+            };
+            scratch_string
+        };
+        let ustack_rep = format_native_stack(self.ustack);
+        let kstack_rep = format_native_stack(self.kstack);
+        let final_rep = write!(
+            f,
+            "RawAggregatedSample:\npid: {}\nustack: {}\nkstack: {}\ncount: {}",
```
javierhonduco, line 162: Same as below, feel free to use the formatting helpers 😄

src/main.rs
```diff
@@ -202,18 +202,64 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                Err(_) => (
+                    "<could not fetch proc comm>".to_string(),
+                    "<could not fetch thread name>".to_string(),
+                ),
             };
 
             writeln!(
                 folded,
-                "{:?}{}{} {}",
+                "{};{}{}{} {}",
```
javierhonduco, line 260: no more quotes!!! 😄

src/main.rs
```diff
@@ -202,18 +202,64 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                    Err(_) => (
+                        "<could not fetch proc stat>".to_string(),
+                        "<could not fetch thread name>".to_string(),
+                    ),
                 },
-                Err(_) => "<could not get proc comm>".to_string(),
+                // Completely failed to look up the PID/TID in procfs
+                Err(_) => (
+                    "<could not fetch proc comm>".to_string(),
+                    "<could not fetch thread name>".to_string(),
```
javierhonduco, line 254: nit: should we call them the same, either both name or comm?

javierhonduco (approved):
Looks great! 🎉  Preemptively approving, I've left some small comments and nits.

As we discussed, we are going to move most of the thread name infrastructure to the metadata providers once we focus on that area. This will reduce the amount of reads from procfs.

As we don't have integration tests yet it would be super helpful if you could add some small "test plan" to the PR to show how this looks!

javierhonduco:
(oh and let's squash the commits when merging!)

javierhonduco:
Good stuff!! 🎉

PR #32, opened 2024-05-07
build: Disable debuginfod support in readelf

javierhonduco:
It pulls a bunch of code we don't need, such as a web server.

Test Plan
=========

CI

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,7 @@
           pkgs = import nixpkgs {
             inherit system overlays;
           };
-          elfutils-without-zstd = pkgs.elfutils.overrideAttrs (attrs: {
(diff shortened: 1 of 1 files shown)


gmarler:
The use of `'` to denote that you're using an altered configuration of elfutils confused me for a bit - finally find a reference to the naming convention and what it denotes [here](https://www.reddit.com/r/NixOS/comments/ttaw5u/what_is_the_purpose_of_single_quotes_after/).

But it would be good to have a comment before such lines denoting that you're creating elfutils "prime" - or if we end up doing this a lot (which we probably will), put a blurb in the project README and reference that in a comment.

This is the first time I've ever heard of this practice, but it's certainly a good one.

javierhonduco:
Good point, will add this in a style guide / contributing doc in few weeks :)

PR #34, opened 2024-05-07
Update flake

javierhonduco:
And fix lint errors. Seems like .cargo/config now lives in .cargo/config.toml


javierhonduco:
The way the kernel images are fetched is quite kludgy, we need to rework this, and this is blocked until we use the derivation paths rather than the hardcoded hashes

src/profiler.rs
```diff
@@ -930,7 +930,7 @@ impl Profiler<'_> {
             return true;
         }
 
-        return self.filter_pids.get(&pid).is_some();
+        self.filter_pids.contains_key(&pid)
```
gmarler, line 933: This does look nicer/cleaner

javierhonduco:
cc @gmarler I've made some small changes to the CLI arguments. To run the kernel tests `nix run .#vmtest`

PR #35, opened 2024-05-07
ci: Build container image

javierhonduco:
Test Plan
=========

CI


javierhonduco:
Mmm seems like cancellation of duplicated workflows is not working properly...

PR #39, opened 2024-05-10
Add get_online_cpus() utiilty function

gmarler:
Attaching perf events can only be done for CPUs that are online - merely being present or possible isn't sufficient to allow such an event to be attached.

It's also possible for online CPU ranges to not be contiguous, due to NUMA topology or CPUs being offlined in some fashion.

So adding a utility function that provides a Vec of individual (not ranges of) online CPUs and using that instead of `num_possible_cpus()`, which could be much larger than the number of online CPUs.

Also updating code that tries to size certain data structures based on the number of online CPUs.

There are still [places](https://github.com/javierhonduco/lightswitch/blob/c89f30fa7b6dcddb68c9a051d516e4d9e98bc6eb/src/profiler.rs#L473) where the possible number of CPUs is still needed.

This is due to certain eBPF (libbpf) internals that have to allocate buffers based on the possible number of CPUs, rather than the online number - you'll only notice this on machines where they differ, which is pretty often for both physical and virtual machines.

Tested this one line of code on a host where the online number of CPUs was 4, but the possible was 8 to verify it works as expected (and fails when you try to use the online number instead).

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,4 +1,4 @@
-use anyhow::Error;
+use anyhow::{Context, Error};
 use std::{fs::File, io::Read};
 
 #[derive(Debug, PartialEq)]
@@ -49,17 +49,17 @@ fn _read_cpu_range(path: &str) -> Result<Vec<u32>, Error> {
                 cpu_range
                     .trim_end()
                     .parse::<u32>()
-                    .map_err(|e| format!("Failed to parse lone CPU: {:?}", e))?,
+                    .with_context(|| "Failed to parse lone CPU".to_string())?,
             ),
             Some(index) => {
                 let start = cpu_range[..index]
                     .trim_end()
                     .parse::<u32>()
-                    .map_err(|e| format!("Failed to parse starting CPU: {:?}", e))?;
+                    .with_context(|| "Failed to parse starting CPU".to_string())?;
                 let end = cpu_range[index + 1..]
                     .trim_end()
                     .parse::<u32>()
-                    .map_err(|e| format!("Failed to parse ending CPU: {:?}", e))?;
+                    .with_context(|| "Failed to parse ending CPU".to_string())?;
                 cpus.extend(start..end + 1);
             }
         }
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,3 +1,6 @@
+use anyhow::Error;
+use std::{fs::File, io::Read};
+
 #[derive(Debug, PartialEq)]
 pub struct AddressBlockRange {
     pub addr: u64,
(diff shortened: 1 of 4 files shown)


src/util.rs
```diff
@@ -33,9 +36,48 @@ pub fn summarize_address_range(low: u64, high: u64) -> Vec<AddressBlockRange> {
...
+    fh.read_to_string(&mut cpu_range_str)?;
+
+    for cpu_range in cpu_range_str.split(',') {
+        let rangeop_result = cpu_range.find('-');
+        match rangeop_result {
+            None => cpus.push(
+                cpu_range
+                    .trim_end()
+                    .parse::<u32>()
+                    .expect("Unable to parse lone CPU number"),
```
javierhonduco, line 52: How do you feel about returning an error here rather than panicking?

src/util.rs
```diff
@@ -33,9 +36,48 @@ pub fn summarize_address_range(low: u64, high: u64) -> Vec<AddressBlockRange> {
...
+            ),
+            Some(index) => {
+                let start = cpu_range[..index]
+                    .trim_end()
+                    .parse::<u32>()
+                    .expect("starting CPU number");
+                let end = cpu_range[index + 1..]
+                    .trim_end()
+                    .parse::<u32>()
+                    .expect("ending CPU number");
```
javierhonduco, line 62: Same here and above

javierhonduco (commented):
Thanks for tackling this! Left some comments (might be slow to reply today, not feeling very well)

src/util.rs
gmarler: That was the intent - fixing that.

src/util.rs
gmarler: ditto

src/util.rs
gmarler: Fixed.

src/util.rs
gmarler: Fixed

src/util.rs
```diff
@@ -155,4 +197,50 @@ mod tests {
...
+        assert_eq!(cpus, expected);
+
+        tmp_file.rewind().unwrap();
+        writeln!(tmp_file, "0-1,3,7-9,48,49").unwrap();
+
+        let cpus = _read_cpu_range(file_str).unwrap();
+        assert_eq!(
+            cpus,
+            (0..=1)
+                .chain(3..=3)
```
javierhonduco, line 236: TIL about `chain()`, very cool!

javierhonduco (approved):
Thanks for this very needed change! LGTM. 

I thought I had left a comment on some small improvement to how we test this. Ideally we would not rely on the filesystem (it might be wiped during test execution and cause the test to fail), which perhaps we could address by passing a memory buffer rather than an actual file (see how we do this for the kernel symbols parser). This is not a high priority of course, so I am happy to merge as-is and change this later on!

javierhonduco:
BTW I tested it locally on my x86 box and worked like a charm 😄

PR #44, opened 2024-05-20
Improve main executable detection

javierhonduco:
The logic here is a bit flimsy and should be reworked, but let's at least mark first executables as the "main" ones.

More context in the comment but overall this is not always correct, in particular Rust + static musl is broken due how the loader operates for that case.

We should understand the kernel and loader's logic in detail and fully replicate it here.

Test Plan
=========

Ran locally for several minutes without issues. The profilers looked good (see PR) and 0 unexpected errors were emitted.

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -47,7 +47,7 @@ impl Collector {
                     load_offset: v.load_offset,
                     load_vaddr: v.load_vaddr,
                     is_dyn: v.is_dyn,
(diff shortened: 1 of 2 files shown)


javierhonduco:
cc @gmarler. I will spend some time when I have the cycles improving this because it's not 100% correct in some situations, I don't think this is standardized anywhere so reading the kernel / loader sources might be the only way to do this properly

PR #45, opened 2024-05-28
Implement unwind information reset

javierhonduco:
This implmementation has several shortcomings, and doesn't have any rate-limiting in place. We will continue to improve this and have plans to revisit how we store chunks (as this can be optimized) as well as how the unwind information is laid out to try to prevent expensive state resets.

Test Plan
=========

Ran on busy hosts that used all shards. The state was resest and profiling continued as usual.

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -89,6 +89,18 @@ pub struct NativeUnwindState {
     shard_index: u64,
 }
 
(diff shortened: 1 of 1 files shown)


javierhonduco:
cc @gmarler as this should fix the bug you reported. Let me know if that's not the case!

PR #54, opened 2024-08-08
Allow eBPF map sizes and general buffer size for profiler to be changed via CLI options

gmarler:
This PR adds several CLI options to set defaults for and allow modifying the size of most eBPF maps and the perf buffer for the profiler:

* `--mapsize-info` prints the eBPF map sizes after they've been created
* `--mapsize-stacks` to adjust max number of individual stacks to capture before they are aggregated
* `--mapsize-aggregated-stacks` to adjust the max number of unique stacks to be stored after aggregation from the stacks map
* `--mapsize-unwind-info-chunks` to adjust the max number of chunks allowed to be in a single shard
* `--mapsize-unwind-tables` to adjust the max number of shards allowed
* `--mapsize-rate-limits` to adjust when to impose rate limits on incoming profiling data
* `--perf-buffer-bytes` to adjust the size of each profiler perf buffer

Also added a config validator for `--perf-buffer-bytes`, as the value must be a power of 2.

Added tests for the validator function.

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -127,7 +127,7 @@ impl AggregatorCollector {
     }
 }
 
-/// Aggregagates the samples in memory, which might be acceptable when profiling for short amounts of time.
+/// Aggregates the samples in memory, which might be acceptable when profiling for short amounts of time.
 impl Collector for AggregatorCollector {
     fn collect(
         &mut self,
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,11 +81,7 @@ fn value_is_power_of_two(s: &str) -> Result<usize, String> {
 fn is_power_of_two(v: usize) -> bool {
     // NOTE: Neither 0 nor 1 are a power of 2 (ignoring 2^0 for this use case),
     //       so rule them out
-    if (v != 0) && (v != 1) && ((v & (v - 1)) == 0) {
-        true
-    } else {
-        false
-    }
+    (v != 0) && (v != 1) && ((v & (v - 1)) == 0)
 }
 
 /// Given a non-prime unsigned int, return the prime number that precedes it
--- a/src/main.rs
+++ b/src/main.rs
@@ -63,17 +63,28 @@ fn sample_freq_in_range(s: &str) -> Result<u16, String> {
     Ok(sample_freq as u16)
 }
 
-// Return a value if it's a power of 2, otherwise Error
+// Clap value_parser() in the form of: Fn(&str) -> Result<T,E>
+// Convert a &str into a usize, if possible, and return the result if it's a
+// power of 2, otherwise Error
 fn value_is_power_of_two(s: &str) -> Result<usize, String> {
     let value: usize = s
         .parse()
         .map_err(|_| format!("`{s}' isn't a valid usize"))?;
     // Now we have a value, test whether it's a power of 2
-    // NOTE: Neither 0 nor 1 are a power of 2, so rule them out
-    if (value != 0) && (value != 1) && ((value & (value - 1)) == 0) {
+    if is_power_of_two(value) {
         Ok(value)
     } else {
         Err(format!("{} is not a power of 2", value))
+    }
+}
+
+fn is_power_of_two(v: usize) -> bool {
+    // NOTE: Neither 0 nor 1 are a power of 2 (ignoring 2^0 for this use case),
+    //       so rule them out
+    if (v != 0) && (v != 1) && ((v & (v - 1)) == 0) {
+        true
+    } else {
+        false
     }
 }
 
@@ -375,7 +386,7 @@ mod tests {
     use assert_cmd::Command;
     use clap::Parser;
     use rand::distributions::{Distribution, Uniform};
-    use rstest::rstest;
+    use rstest::{fixture, rstest};
     use std::collections::HashSet;
 
     #[test]
@@ -466,27 +477,37 @@ mod tests {
         }
     }
 
-    #[rstest]
-    fn should_be_powers_of_two() {
-        let mut test_uint_strings = vec![];
-        for shift in 1..63 {
+    // Powers of 2 in usize range
+    #[fixture]
+    fn power_of_two_usize() -> Vec<usize> {
+        let mut test_usizes = vec![];
+        for shift in 0..63 {
             let val: usize = 2 << shift;
+            test_usizes.push(val);
+        }
+        test_usizes
+    }
+
+    // Powers of 2 represented as Strings
+    #[fixture]
+    fn power_of_two_strings(power_of_two_usize: Vec<usize>) -> Vec<String> {
+        let mut test_uint_strings = vec![];
+        for val in power_of_two_usize {
             let val_str = val.to_string();
             test_uint_strings.push(val_str);
         }
-        for val_string in test_uint_strings {
-            assert!(value_is_power_of_two(val_string.as_str()).is_ok())
-        }
+        test_uint_strings
     }
 
-    #[rstest]
-    fn should_not_be_powers_of_two() {
-        let mut test_uint_stringset: HashSet<String> = HashSet::new();
+    // This fixture produces 5 million random results from the range of usize
+    // integers that are NOT powers of 2
+    #[fixture]
+    fn all_but_power_of_two_usize(power_of_two_usize: Vec<usize>) -> Vec<usize> {
+        let mut test_usize_set: HashSet<usize> = HashSet::new();
+        let mut test_usize_not_p2: Vec<usize> = vec![];
         // usizes that ARE powers of two, for later exclusion
-        for shift in 0..63 {
-            let val: usize = 2 << shift;
-            let val_string = val.to_string();
-            test_uint_stringset.insert(val_string);
+        for val in power_of_two_usize {
+            test_usize_set.insert(val);
         }
         // Now, for a random sampling of 500000 integers in the range of usize,
         // excluding any that are known to be powers of 2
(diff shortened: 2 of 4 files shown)


src/collector.rs
```diff
@@ -127,7 +127,7 @@ impl AggregatorCollector {
     }
 }
 
-/// Aggregagates the samples in memory, which might be acceptable when profiling for short amounts of time.
+/// Aggregates the samples in memory, which might be acceptable when profiling for short amounts of time.
```
javierhonduco, line 130: 🦅 👀

src/main.rs
```diff
@@ -61,6 +62,20 @@ fn sample_freq_in_range(s: &str) -> Result<u16, String> {
     Ok(sample_freq as u16)
 }
 
+// Return a value if it's a power of 2, otherwise Error
+fn value_is_power_of_two(s: &str) -> Result<usize, String> {
+    let value: usize = s
+        .parse()
+        .map_err(|_| format!("`{s}' isn't a valid usize"))?;
+    // Now we have a value, test whether it's a power of 2
+    // NOTE: Neither 0 nor 1 are a power of 2, so rule them out
```
javierhonduco, line 71: I think we should only exclude 0: `2ˆ0 == 1`

src/main.rs
```diff
@@ -384,4 +455,42 @@ mod tests {
             }
         }
     }
+
+    #[rstest]
+    fn should_be_powers_of_two() {
+        let mut test_uint_strings: Vec<String> = vec![];
```
javierhonduco, line 461: nit: no need to specify the type here, in lightswitch we tend to avoid doing this explicitly unless the type checker can't infer the type

src/main.rs
```diff
@@ -384,4 +455,42 @@ mod tests {
...
+        }
+        // Now, for a random sampling of 500000 integers in the range of usize,
+        // excluding any that are known to be powers of 2
+        let between = Uniform::from(0..=usize::MAX);
+        let mut rng = rand::thread_rng();
+        for _ in 0..500000 {
+            let usize_int: usize = between.sample(&mut rng);
+            let usize_int_string = usize_int.to_string();
+            if test_uint_stringset.contains(&usize_int_string) {
+                println!("{}", usize_int_string);
```
javierhonduco, line 489: nit: if we want to leave this for latter debugging, let's add a bit more context

src/profiler.rs
```diff
@@ -163,6 +163,8 @@ pub struct Profiler<'bpf> {
     duration: Duration,
     // Per-CPU Sampling Frequency of this profile in Hz
     sample_freq: u16,
+    // Size of each perf buffer, in include_bytes!
```
javierhonduco, line 166:
```suggestion
    // Size of each perf buffer, in bytes
```

javierhonduco: Daft autocomplete! 😄

src/profiler.rs
```diff
@@ -285,34 +283,111 @@ impl fmt::Display for SymbolizedAggregatedSample {
...
+    pub perf_buffer_bytes: usize,
+    pub mapsize_info: bool,
+    pub mapsize_stacks: u32,
+    pub mapsize_aggregated_stacks: u32,
+    pub mapsize_unwind_info_chunks: u32,
+    pub mapsize_unwind_tables: u32,
+    pub mapsize_rate_limits: u32,
 }
 
+// impl Default for Profiler<'_> {
```
javierhonduco, line 300: Is this intended?

src/profiler.rs
```diff
@@ -395,23 +471,25 @@ impl Profiler<'_> {
...
         let chan_send = self.chan_send.clone();
         let perf_buffer = PerfBufferBuilder::new(self.bpf.maps().events())
-            .pages(PERF_BUFFER_BYTES / page_size::get())
+            .pages(self.perf_buffer_bytes / page_size::get())
             .sample_cb(move |_cpu: i32, data: &[u8]| {
                 Self::handle_event(&chan_send, data);
             })
             .lost_cb(Self::handle_lost_events)
             .build()
+            // TODO: Instead of unwrap, consume and emit any error - usually
```
javierhonduco, line 480: This should not happen due to the validation in the CLI, right? It's true that as a library it's a possible error, though

src/profiler.rs
```diff
@@ -425,6 +503,8 @@ impl Profiler<'_> {
                     warn!("lost {} events from the tracers", lost_count);
                 })
                 .build()
+                // TODO: Instead of unwrap, consume and emit any error - usually
```
javierhonduco, line 506: Same as above

src/main.rs
```diff
@@ -61,6 +62,20 @@ fn sample_freq_in_range(s: &str) -> Result<u16, String> {
     Ok(sample_freq as u16)
 }
 
+// Return a value if it's a power of 2, otherwise Error
+fn value_is_power_of_two(s: &str) -> Result<usize, String> {
```
javierhonduco, line 66: I find a bit odd that the argument is a string, I would have expected some integer here

src/main.rs
gmarler: Since this is a Clap validator routine, I'll check whether allowing the value of 1 is accepted by eBPF - if so, then I'll let it pass with a note in the validation routine - otherwise, I *have* to exclude it to prevent crashes.

src/main.rs
gmarler: As long as the type is inferred, I have no objection - will try it out.

src/main.rs
gmarler: No, this was intended to be removed - will elide.

src/main.rs
javierhonduco: That's fair, in case you leave that as-is we should update the comment, though

gmarler: Yeah, have to exclude 1, or you get this:

```
$ cargo run -- --perf-buffer-bytes 1 -D 1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.12s
     Running `sudo -E target/debug/lightswitch --perf-buffer-bytes 1 -D 1`
...
thread 'main' panicked at src/profiler.rs:482:14:
called `Result::unwrap()` on an `Err` value: Error: Page count must be power of two
```

src/main.rs
gmarler: Yes, works just fine without type.

src/profiler.rs
gmarler: Done!

src/profiler.rs
gmarler: It was intended to remove the Default impl for Profiler and replace it with a Default impl for ProfilerConfig - will do that now.

src/profiler.rs
gmarler: Yes, I added this comment before I added the validator routine for Clap - but there are other possible ways this could fail.  I suppose just panic'ing if we hit one of those is just as good, because you'll at least get the error message saying _why_ the panic occurred.
Just don't like seeing naked `unwrap()`s

src/main.rs
gmarler: Clap validator routines all take `&str`s - the conversion to an integer type happens inside Clap itself - caught me by surprise the first time as well.

javierhonduco: Makes sense! In that case I think it would be nicer to have this function to accept an integer and the validator to accept a string reference, convert it to the right type, and pass it to the power-of-two-validator function. What do you think?

src/profiler.rs
javierhonduco: haha fair, it's a hard exception, so perhaps I can replace them with `expect` to add more context, what do you say?

src/main.rs
javierhonduco: 👀  Seems like the kernel validator is being funny here! Good to know, even though I don't think we would ever choose a size that small 😄

src/profiler.rs
gmarler: Perfectly fine.

src/main.rs
gmarler: Should have been more clear - these function types always take strings, and have to do the conversions themselves internally to whatever type you want to end up with - in this case, an integer.

Haven't seen an example where you can separate the concepts.

javierhonduco: You were clear before, but I meant to separate the pure numeric power-of-two-validator so you can test it without string conversions, and also because it makes more sense for this function to take an integer. As clap requires validators to receive strings, we can have another function that converts string -> int and then performs the check

src/profiler.rs
```diff
@@ -288,37 +286,135 @@ impl fmt::Display for SymbolizedAggregatedSample {
...
+// with the defaults defined here.
+impl Default for ProfilerConfig {
+    fn default() -> Self {
+        Self {
+            libbpf_debug: false,
+            bpf_logging: false,
+            duration: Duration::MAX,
+            sample_freq: 0,
+            perf_buffer_bytes: 0,
+            mapsize_info: false,
```
javierhonduco, line 315: Not sure how I feel about a default configuration that would prevent the profiler from working 🤔

javierhonduco: I understand the intent, but this doesn't feel right, wondering if we could do this differently

tests/integration_test.rs
```diff
@@ -99,14 +99,22 @@ fn test_integration() {
...
+        bpf_logging: bpf_test_debug,      // changed from default
+        duration: Duration::from_secs(5), // changed from default
+        sample_freq: 999,                 // changed from default
+        perf_buffer_bytes: 512 * 1024,
+        mapsize_info: false,
+        mapsize_stacks: 100000,
+        mapsize_aggregated_stacks: 10000,
+        mapsize_unwind_info_chunks: 5000,
+        mapsize_unwind_tables: 65,
+        mapsize_rate_limits: 5000,
```
javierhonduco, line 114: This would be a good place where to pass a good `::default()` I think to fill the other fields we aren't setting

tests/integration_test.rs
```diff
@@ -99,14 +99,22 @@ fn test_integration() {
     let collector = Arc::new(Mutex::new(
         Box::new(AggregatorCollector::new()) as Box<dyn Collector + Send>
     ));
+
+    let profiler_config = ProfilerConfig {
+        libbpf_debug: bpf_test_debug,     // changed from default
+        bpf_logging: bpf_test_debug,      // changed from default
+        duration: Duration::from_secs(5), // changed from default
+        sample_freq: 999,                 // changed from default
```
javierhonduco, line 107: nit: these comments might get out of sync very quickly, perhaps it would be better to remove them

src/profiler.rs
gmarler: I've altered the Default for ProfilerConfig so that it's sane by default, just added a comment that notes it'll almost always be overridden by the Clap config defaults.

tests/integration_test.rs
gmarler: That was the original intent, but forgot - done.

tests/integration_test.rs
gmarler: And.... they're gone!

src/main.rs
gmarler: I've separated the parsing of the string into a usize from validation of the usize being a power of two or not, and added several test fixtures to simplify the tests of both strings and usizes.

PR #60, opened 2024-08-22
Add statistics for encounters of vDSO or JIT

gmarler:
This PR adds counters for two types of frames that may be encountered during unwinding that we don't currently _handle_, but want to know if we've encountered them on particular workloads:

* vDSO
* JIT

We also stop clearing the statistics records on each collection interval, so that we get a good overall set of statistics for run lifetimes of the agent process.


javierhonduco:
LGTM, thanks!

PR #61, opened 2024-08-29
Chore: Add benchmark for `KsymIter::from_kallsyms()`

patnebe:
## Context
* Some performance improvements to `KsymIter`

## Changes
* Added the first benchmark
* Reduce the number of heap allocations in `KsymIter::from_kallsyms()` by 
  * replacing `lines()` with `read_line()` + a reusable buffer.
  * using iterators directly instead of collecting the result of `line.split()` into a vector.

## Test Plan
* Unit tests
* Benchmark results on a test kallsyms file containing 1013568 lines. I consistently observed similar improvements for a file with 675712 entries. However, it's worth noting that I don't see the same performance gains when I run this against a file with 337856 entries.
```md
Benchmarking Read /proc/kallsyms/Measure read all ksyms: Warming up for 5.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.4s, or reduce sample count to 50.
Read /proc/kallsyms/Measure read all ksyms
                        time:   [84.522 ms 84.705 ms 84.871 ms]
                        change: [-24.298% -24.049% -23.840%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) low severe
  5 (5.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe
```


src/ksym.rs
```diff
@@ -33,23 +35,28 @@ impl<R: Read> Iterator for KsymIter<R> {
...
-                        let symbol_name = v[2];
-
-                        let current = Ksym {
-                            start_addr: start_addr.unwrap(),
-                            symbol_name: symbol_name.to_string(),
-                        };
+            let buffer = &mut self.line;
+            buffer.clear();
+            match self.file.read_line(buffer) {
+                Ok(bytes_read) => {
```
javierhonduco, line 41: style nit, we can use Rust's powerful pattern matching here to make this clearer:

```rust
match match self.file.read_line(buffer) { 
  Ok(0) => {
    // End of file
    return None;
  },
  Ok(bytes_read) => { ... }
}
```

patnebe: Updated this. Also aded a note to the PR description about the results I see when I test on different kallsyms file sizes.

Cargo.toml
```diff
@@ -53,3 +54,7 @@ glob = "0.3.1"
 
 [profile.dev.package."*"]
 opt-level = 3
+
+[[bench]]
+name ="benchmark"
```
javierhonduco, line 59: style nit (not sure why `cargo fmt` doesn't format the Cargo toml files)
```suggestion
name = "benchmark"
```

javierhonduco (approved):
LGTM, great work! Left just a small nit

src/ksym.rs
javierhonduco: Thanks!

PR #62, opened 2024-09-02
Fix race while copying `ObjectFileInfo`

javierhonduco:
There was a fundamental race that could (and did!) happen while copying / cloning `ObjectFileInfo`, as we tried to open the file using the original file path, if the file had been removed, re-opening would fail. This commit leverages procfs to reopen a file that might already have been deleted, as long as there are some file descriptors pointing to it, as the kernel won't effectively remove the file until its reference count reaches zero.

Test Plan
=========

Added test to exercise this code path. Also run the profiler while compiling the Linux kernel as this as this triggered the crash in the past.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,6 +1288,7 @@ dependencies = [
  "ring",
  "rstest",
  "tempdir",
+ "tempfile",
  "thiserror",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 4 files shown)


javierhonduco:
Note that we don't use this re-opening technique yet for userspace symbolization, so that could still fail if the file is deleted at that point

PR #63, opened 2024-09-05
metadata: Extract facility to find process and thread names

javierhonduco:
The previous implementation was buggy. This commit fixes this and adds tests to this new logic.

Test Plan
=========

Added new tests + manual tests

**current**

image

**with this PR**

image


javierhonduco:
cc/ @gmarler

gmarler (approved):
Yes, this is a *much* better implementation - like the `TaskName::errored()` in particular.

PR #64, opened 2024-09-05
usym: Fix order when dealing with inlined functions

javierhonduco:
Test Plan
=========

Added tests to ensure we return frames in the right order, and catch potential regressions with blazesym. Also removed the addr2line backend as it is not complete and it's not adapted to the new inteface. I still think there's value to have it so will re-add it down the line.


javierhonduco:
cc @gmarler @patnebe

PR #70, opened 2024-09-17
Detect system info + available BPF features on startup

patnebe:
## Context
* Early detection of system info and BPF features to ensure the required runtime dependencies are met.
* This information will also provide a better UX through improved error reporting on startup.

## Changes
* Add a new BPF program to detect what BPF features are available on the host.
* Check that `procfs` and `tracefs` are mounted.
* Check support for `tracepoints`, `perf_events`, and a few bpf helpers.

## Test Plan
* Manual tests + Unit(? / integration?) tests.


src/system_info.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub os_release: String,
+    pub procfs_mount_detected: bool,
+    pub tracefs_mount_detected: bool,
+    pub tracepoints_support_detected: bool,
+    pub software_perfevents_support_detected: bool,
+    pub available_bpf_features: BpfFeatures,
+}
+
+fn tracefs_mount_detected() -> bool {
+    return Path::new(PROCFS_PATH).exists();
```
javierhonduco, line 39: nice! very clean

src/system_info.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+fn get_trace_sched_event_id(trace_event: &str) -> Result<u32> {
+    if !tracefs_mount_detected() {
+        return Err(anyhow!("Failed to detect tracefs"));
```
javierhonduco, line 44: We should probably have custom errors. I appreciate that currently the project abuses `anyhow`, and we should change this, but let's try to make this errors into their own variants of an enum. Happy to send some hints if you need them!

src/system_info.rs
```diff
@@ -0,0 +1,211 @@
...
+    match read_to_string(path) {
+        Ok(id) => match id.trim().parse::<u32>() {
+            Ok(val) => Ok(val),
+            Err(err) => Err(anyhow!(
+                "Failed to read event={} id, err={}",
+                trace_event,
+                err
+            )),
+        },
+        Err(_) => Err(anyhow!("Failed to read event={} id", trace_event)),
```
javierhonduco, line 62: Using [`or_else`](https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else)  + `?` might simply the code

src/system_info.rs
```diff
@@ -0,0 +1,211 @@
...
+                }
+                true
+            }
+            Err(_) => false,
+        }
+    }
+}
+
+fn tracepoints_detected() -> bool {
+    let mut tracepoints_supported = true;
```
javierhonduco, line 81: style nit: not need for this variable, feel free to just return either `true` or `false` directly

src/system_info.rs
```diff
@@ -0,0 +1,211 @@
...
+            0,  /* flags */
+        )
+    };
+
+    if fd < 0 {
+        tracepoints_supported = false;
+        return tracepoints_supported;
+    }
+
+    if unsafe { close(fd) } != 0 {
```
javierhonduco, line 112: We can look into wrapping this into a type that on `drop()` attempts to close this. The `Drop` trait is akin to RAII in C++

javierhonduco (commented):
Looks very good overall!! Great progress so far! 

We might want to look into the kernel test errors, and another thing we could do was moving all this code that is pretty much self-contained (and potentially) reusable to a top level crate similarly to how `lightswitch-proto` is implemented. What do you think?

src/system_info.rs
patnebe: Fair point. Added a few custom errors to this file

patnebe:
> We might want to look into the kernel test errors, 

FWIW, the error seemed to be related to the unavailability of the `hrtimer_start_range_ns` kprobe (or maybe kprobes in general on the VM?). Perhaps some kernel flag is missing( ?). 

Instead of spending time figuring out how to get that to work, I decided to hook the BPF feature detection program to a tracepoint. Main reason being we already rely on tracepoints and currently don't use kprobes anywhere else.

> another thing we could do was moving all this code that is pretty much self-contained (and potentially) reusable to a top level crate similarly to how `lightswitch-proto` is implemented. What do you think?

That sounds reasonable. Maybe `lightswithc_sys_probe` would be a good name for this crate? Open to suggestions :)

lightswitch-sys-probe/src/system_info.rs
```diff
@@ -0,0 +1,249 @@
...
+                0,  /* flags */
+            ) as c_int
+        },
+    };
+
+    if fd.fd < 0 {
+        return false;
+    }
+
+    true
```
patnebe, line 153: fd.fd >= 0

lightswitch-sys-probe/src/system_info.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    pub fn has_minimal_requirements(&self) -> bool {
+        let bpf_features = &self.available_bpf_features;
+        self.tracefs_mount_detected
+            && self.procfs_mount_detected
+            && self.software_perfevents_support_detected
+            && self.tracepoints_support_detected
+            && bpf_features.can_load_trivial_bpf_program
+            && bpf_features.has_ring_buf
```
javierhonduco, line 215: We don't use ring buffers now, and when we do, most likely it will be dynamically chosen depending on the availability of ring buffers

lightswitch-sys-probe/src/bpf/vmlinux.h
```diff
@@ -0,0 +1,7 @@
+#ifdef __TARGET_ARCH_x86
+#include "vmlinux_x86.h"
```
javierhonduco, line 2: Given how large vmlinux is and that we might want to keep just on copy, what do you think of using either symlinks or including the file from the other crate directly here?

javierhonduco:
> FWIW, the error seemed to be related to the unavailability of the hrtimer_start_range_ns kprobe (or maybe kprobes in general on the VM?). Perhaps some kernel flag is missing(?). So instead of spending time figuring out how to get that to work, I decided to hook the BPF feature detection program to a tracepoint instead. The main reason is that we already rely on tracepoints and currently don't use kprobes anywhere else.


Makes sense!

> That sounds reasonable. Maybe lightswitch_sys_probe would be a good name for this crate? Open to suggestions :)

What about `lightswitch-features-probing` / `lightswitch-capabilities` or something like this? `-sys` is typically used for C bindings

lightswitch-sys-probe/src/bpf/vmlinux.h
patnebe: Done. Changed this to a symlink

lightswitch-sys-probe/src/system_info.rs
patnebe: Right. I'm relaxed the ringbuf requirement.

lightswitch-capabilities/src/system_info.rs
```diff
@@ -0,0 +1,244 @@
...
+        self.tracefs_mount_detected
+            && self.procfs_mount_detected
+            && self.software_perfevents_support_detected
+            && self.tracepoints_support_detected
+            && bpf_features.can_load_trivial_bpf_program
+            && bpf_features.has_tail_call
+    }
+}
+
+// TODO: How can we make this an integration/system test?
```
javierhonduco, line 219: It's totally fine to leave it here. Personally I don't think the difference between purely no-IO unittests and integration tests matters that much, unless we need to test a binary, in that case we should use the tool level folder instead. The standard in the Rust world is to add integration tests in `test/` so here would go on `lightswitch-capabilities/test`.

javierhonduco (approved):
Great job! LGTM

PR #74, opened 2024-09-18
Exclude profiler

gmarler:
Add option `--exclude-self` that will prevent profiling the profiler.

Updated test for the new option.


src/profiler.rs
```diff
@@ -1216,6 +1222,10 @@ impl Profiler<'_> {
     }
 
     fn should_profile(&self, pid: i32) -> bool {
+        if self.exclude_self && (pid == std::process::id().try_into().unwrap()) {
```
javierhonduco, line 1225: nit: no need for the parenthesis, and as we know that the type fits, we can use `as` until we have our own type

```suggestion
        if self.exclude_self && pid == std::process::id() as i32 {
```

src/profiler.rs
```diff
@@ -398,6 +402,7 @@ impl Profiler<'_> {
             .lightswitch_config
             .verbose_logging
             .write(profiler_config.bpf_logging);
+        let exclude_self = profiler_config.exclude_self;
```
javierhonduco, line 405: nit: no need to extract to a variable if we only use it here

javierhonduco (approved):
LGTM! Left two small nits

src/profiler.rs
gmarler: We do use it in the Profiler initialization

src/profiler.rs
gmarler: Fixed

PR #83, opened 2024-10-08
Pluggable metadata library for generating sample labels

patnebe:
## Context
* Collection of various off-CPU metadata for profiles to provide additional context during analysis. These metadata are applied as labels to each pprof sample, and will be used to enable rich filtering of profiles on the backend UI.

## Changes
* Define a `MetadataProvider` trait and a `GlobalMetadataProvider` to encapsulate the logic for retrieving task metadata.
* Move the existing labels into the provider, and hook up the provider in main.
* Implement basic LRU caching to limit the number of processes for which labels are held in memory.

## Future changes for a follow-up PR
* Runtime detection
* Mapping k8s units of work to tasks

## Test Plan
* Manual tests + Unit tests.


lightswitch-metadata-provider/src/metadata_provider.rs
```diff
@@ -0,0 +1,156 @@
...
+pub trait MetadataProvider {
+    /// Return a vector of labels for the provided task id.
+    /// Labels returned by this function will be assumed to apply
+    /// to all task_ids in the same process/tgid as the provided task_id.
+    fn get_metadata(&self, task_id: i32) -> Result<Vec<Label>, MetadataProviderError>;
+}
+pub type ThreadSafeMetadataProvider = Arc<Mutex<Box<dyn MetadataProvider + Send>>>;
+
+pub struct GlobalMetadataProvider {
+    pid_label_cache: LruCache</*pid*/ i32, Vec<Label>>,
```
patnebe, line 28: I think might be worth deduplicating these labels using an interning/string table approach to limit RSS usage. I'm planning on adding that as a follow-up PR to keep this diff minimal.

Unless you think we should be making those changes in this PR? @javierhonduco If so, I'm happy to add those here :)

javierhonduco: Sorry missed this somehow! Doing them in another PR is fine

lightswitch-metadata-provider/src/metadata_provider.rs
```diff
@@ -0,0 +1,156 @@
+use crate::process_metadata::ProcessMetadata;
+use crate::system_metadata::SystemMetadata;
+use crate::taskname::TaskName;
+
+use anyhow::Result;
+use lightswitch_proto::label::{Label, LabelValueStringOrNumber};
```
javierhonduco, line 6: I think it might be better to not rely on any pprof definitions here, there are several reasons I can think of for not relying on them such as:
- clearer boundary definition between the different crates
- not tying that structure to the metadata system, that might use a different 
- not needing a protocol buffers compiler to use the metadata crate

Happy to discuss.

lightswitch-metadata-provider/src/metadata_provider.rs
patnebe: Sounds good!

lightswitch-metadata-provider/src/metadata_provider.rs
patnebe: Yeah, I think these are reasonable arguments against doing this. My initial motivation was to avoid additional translation when attaching the metadata labels to the pprof output. But I now see that this is acceptable given the trade-offs listed above. 

I've updated the diff to prevent coupling between the metadata and proto crates.

lightswitch-metadata-provider/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[dependencies]
+anyhow = "1.0.86"
+lru = "0.12.4"
+nix = { version = "0.29.0", features = ["user", "process"] }
+procfs = "0.16.0"
+tracing = "0.1.40"
+thiserror = "1.0.63"
+
+[build-dependencies]
```
javierhonduco, line 14: We aren't using this section explicitly, perhaps we could remove it?

lightswitch-metadata-provider/src/process_metadata.rs
```diff
@@ -0,0 +1,58 @@
+use crate::metadata_label::{MetadataLabel, MetadataLabelValue};
+
+use std::result::Result::Ok;
```
javierhonduco, line 3: This should be in the scope already, let me know if removing it causes any issues

lightswitch-metadata-provider/src/process_metadata.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Debug, Error)]
+pub enum ProcessMetadataError {
+    #[error("Unable to detect runtime for pid={0}, err={1}")]
+    ErrorDetectingRuntime(i32, String),
+}
+
+impl ProcessMetadata {
+    fn get_runtime(&self, pid: i32) -> Result<Option<String>, ProcessMetadataError> {
+        Err(ProcessMetadataError::ErrorDetectingRuntime(
```
javierhonduco, line 17: There's the [`todo!`](https://doc.rust-lang.org/std/macro.todo.html) macro that is useful in situations like this

lightswitch-metadata-provider/src/system_metadata.rs
```diff
@@ -0,0 +1,82 @@
...
+    format!(
+        "{}:{}",
+        uname.sysname().to_string_lossy(),
+        uname.release().to_string_lossy()
+    )
+}
+
+impl SystemMetadata {
+    pub fn get_metadata(&self) -> Result<Vec<MetadataLabel>, SystemMetadataError> {
+        let uname = match utsname::uname() {
```
javierhonduco, line 25: `map_err` can be useful in these kind of error handling 
```Rust
 let uname = utsname::uname().map_err(|e| SystemMetadataError::ErrorRetrievingSystemInfo(r.desc().to_string()))?;
```

lightswitch-metadata-provider/src/system_metadata.rs
```diff
@@ -0,0 +1,82 @@
...
+            Ok(uname) => uname,
+            Err(err) => {
+                return Err(SystemMetadataError::ErrorRetrievingSystemInfo(
+                    err.desc().to_string(),
+                ));
+            }
+        };
+        let kernel_release_label = MetadataLabel {
```
javierhonduco, line 33: nit (no need to do it now): Some helper constructors might help here, something like `MetadataLabel::from_string_value("kernel_release", get_kernel_release(&uname))`

lightswitch-proto/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 prost = "0.13"
 anyhow = "1.0.89"
+rand = { version = "0.8.5", features = ["small_rng"] }
```
javierhonduco, line 9: Why do we need `rand` as a dependency? 😃

lightswitch-proto/src/profile.rs
```diff
@@ -387,8 +387,8 @@ mod tests {
             pprof.new_label("key", LabelStringOrNumber::String("value".into())),
             pprof.new_label("key", LabelStringOrNumber::Number(123, "pid".into())),
         ];
-        pprof.add_sample(vec![1, 2, 3], 100, labels.clone());
-        pprof.add_sample(vec![1, 2, 3], 100, labels);
+        pprof.add_sample(vec![1, 2, 3], 100, &labels);
```
javierhonduco, line 390: nice

src/profile.rs
```diff
@@ -19,15 +23,26 @@ use crate::profiler::ProcessInfo;
 use crate::profiler::RawAggregatedProfile;
 use crate::usym::symbolize_native_stack_blaze;
 
+/// Converts a metadata provider label type to a proto label type
+fn convert_label_value(metadata_label: MetadataLabelValue) -> LabelStringOrNumber {
```
javierhonduco, line 27: (no need to do this in this PR) Rust has the [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) trait that we could use here

javierhonduco:
This PR didn't break `main`, I did, sorry! Will fix it.

lightswitch-proto/Cargo.toml
patnebe: I probably had that from some code I added to test things. I've now removed it. Thanks!

src/profile.rs
patnebe: Done. Added a wrapper struct to work around the orphan rule.

javierhonduco (approved):
LGTM! Great job! Let's squash all these commits when merging 😄

PR #98, opened 2024-10-29
Move shared crates to workspace

javierhonduco:
Rather than have the same entry multiple times over the different crates, or even worse, accidentally having different versions / features, which would cause more code to be compiled, let's have this in a workspace.dependencies section and refer to it in the various crates.

The build times are pretty much the same

* Before: 72.4s (1m 12.4s)
* After: 73.1s (1m 13.1s)

Test Plan
=========

CI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,51 +11,67 @@ members = [
     "lightswitch-object"
 ]
 
-[dependencies]
-gimli = "0.31.0"
-object = "0.36.4"
+[workspace.dependencies]
 memmap2 = "0.9.5"
-lazy_static = "1.5.0"
 anyhow = "1.0.89"
-thiserror = "1.0.63"
+object = "0.36.4"
 # TODO: Move to a release.
 libbpf-rs = { git = "https://github.com/libbpf/libbpf-rs", rev="4ebf2ac7cd509e9442aff9b9f92164f252adf2a3", features = ["static"] }
-perf-event-open-sys = "4.0.0"
+tracing = "0.1.40"
+thiserror = "1.0.63"
 errno = "0.3.9"
-plain = "0.2.3"
+perf-event-open-sys = "4.0.0"
 procfs = "0.16.0"
+nix = { version = "0.29.0" }
+# workspace dev dependencies have to be defined here
+rand = "0.8.5"
+# workspace build dependencies have to be defined here
+libbpf-cargo = "0.23.3"
+glob = "0.3.1"
+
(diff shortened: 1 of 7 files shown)


javierhonduco:
Generated these with ```cargo build --timings```. Build times did not change much

image

Cargo.toml
```diff
@@ -11,51 +11,67 @@ members = [
...
-lightswitch-capabilities = {path = "./lightswitch-capabilities"}
-lightswitch-object = {path = "./lightswitch-object"}
 ctrlc = "3.4.5"
 crossbeam-channel = "0.5.13"
 libbpf-sys = "1.4.3"
 itertools = "0.13.0"
+lightswitch-metadata-provider = {path = "lightswitch-metadata-provider"}
+lightswitch-proto = { path = "lightswitch-proto"}
+lightswitch-capabilities = {path = "lightswitch-capabilities"}
+lightswitch-object = {path = "lightswitch-object"}
```
rdelfin, line 51: FYI, when you publish you also want to include the version, not just the path (otherwise weird things happen when people try depending on the crate).

flake.nix
```diff
@@ -71,14 +71,17 @@
...
               strace
               gdb
-              # Other tools
+              # Upload container image to registry
               skopeo
+              # Cargo subcommand tools
+              ## To upgrade deps
               cargo-edit
-              # snapshot testing plugin binary
+              ## Snapshot testing
```
rdelfin, line 82: How dare you fix comments

lightswitch-proto/Cargo.toml
```diff
@@ -5,10 +5,13 @@ edition = "2021"
...
 
 [dev-dependencies]
-rand = "0.8.5"
+rand = { workspace = true }
 
 [build-dependencies]
 prost-build = "0.13.3"
+
+[package.metadata.cargo-shear]
+ignored = ["prost"]
```
rdelfin, line 17: lmao what. Why do you need this?

rdelfin (approved):
Couple comments, but generally LGTM, definite improvement

Cargo.toml
javierhonduco: Ohh makes sense! Will change this once I start publishing them!

lightswitch-proto/Cargo.toml
javierhonduco: Ohh good shout, it's from another commit, thanks for noticing!

PR #105, opened 2024-11-04
object: Improve correctness

javierhonduco:
object: Improve correctness
Due to the fact that `object::File` takes a reference to the mmap-ed
file, we needed to do some awful lifetime hacks to fit it in the API we
wanted. While getting rid of the raw pointer and drop impl, Ricardo
noticed that there was some lurking UB too, so this fixes it as well.

Co-authored-by: Ricardo Delfin <me@rdelfin.com>

Test Plan
=========

CI

### DIFF
--- a/lightswitch-object/src/object.rs
+++ b/lightswitch-object/src/object.rs
@@ -34,34 +34,32 @@ pub struct ElfLoad {
 }
 
 #[derive(Debug)]
-pub struct ObjectFile<'a> {
-    leaked_mmap_ptr: *const Mmap,
-    object: object::File<'a>,
+pub struct ObjectFile {
+    /// Warning! `object` must always go above `mmap` to ensure it will be dropped
+    /// before. Rust guarantees that fields are dropped in the order they are defined.
+    object: object::File<'static>, // Its lifetime is tied to the `mmap` below.
+    mmap: Box<Mmap>,
     code_hash: Digest,
 }
 
-impl Drop for ObjectFile<'_> {
-    fn drop(&mut self) {
-        unsafe {
-            let _to_free = Box::from_raw(self.leaked_mmap_ptr as *mut Mmap);
-        }
-    }
-}
-
-impl ObjectFile<'_> {
+impl ObjectFile {
     pub fn new(path: &Path) -> Result<Self> {
         let file = fs::File::open(path)?;
-        let mmap = unsafe { Mmap::map(&file) }?;
-        let mmap = Box::new(mmap);
-        let leaked = Box::leak(mmap);
-        let object = object::File::parse(&**leaked)?;
+        // Rust offers no guarantees on whether a "move" is done virtually or by memcpying,
+        // so to ensure that the memory value is valid we store it in the heap.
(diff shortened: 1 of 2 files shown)


lightswitch-object/src/object.rs
```diff
@@ -35,33 +35,30 @@ pub struct ElfLoad {
...
 impl ObjectFile<'_> {
     pub fn new(path: &Path) -> Result<Self> {
         let file = fs::File::open(path)?;
-        let mmap = unsafe { Mmap::map(&file) }?;
-        let mmap = Box::new(mmap);
-        let leaked = Box::leak(mmap);
-        let object = object::File::parse(&**leaked)?;
+        // Rust offers no guarantees on whether a "move" is done virtually or by memcpying,
+        // so to ensure that the memory value is valid we store it in the heap.
+        let mmap = Box::new(unsafe { Mmap::map(&file) }?);
```
rdelfin, line 50: I would add a separate `// Safety: ` to address [the UB mentioned in memmap2](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#safety). A comment saying the `File` object should never be used. Unsure if you can enforce it with an annotation guaranteeing it goes unused.

rdelfin, line 50: I would add a separate `// Safety: ` to address [the UB mentioned in memmap2](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#safety). A comment saying the `File` object should never be used. Unsure if you can enforce it with an annotation guaranteeing it goes unused.

lightswitch-object/src/object.rs
```diff
@@ -35,33 +35,30 @@ pub struct ElfLoad {
 
 #[derive(Debug)]
 pub struct ObjectFile<'a> {
-    leaked_mmap_ptr: *const Mmap,
+    /// Warning! `object` must always go above `mmap` to ensure it will be dropped
+    /// before. Rust guarantees that fields are dropped in the order they are defined.
     object: object::File<'a>,
```
rdelfin, line 40: Should this lifetime not be:
```suggestion
    object: object::File<'static>,
```

With a comment explaining that this lifetime is attached to the lifetime of `self.mmap`?

rdelfin (commented):
> Co-authored-by: Ricardo Delfin [me@rdelfin.com](mailto:me@rdelfin.com)

Some truly slanderous statements

Just a couple comments

lightswitch-object/src/object.rs
```diff
@@ -34,34 +34,32 @@ pub struct ElfLoad {
 }
 
 #[derive(Debug)]
-pub struct ObjectFile<'a> {
-    leaked_mmap_ptr: *const Mmap,
-    object: object::File<'a>,
+pub struct ObjectFile {
+    /// Warning! `object` must always go above `mmap` to ensure it will be dropped
+    /// before. Rust guarantees that fields are dropped in the order they are defined.
+    object: object::File<'static>, // It's lifetime is tied to the `mmap` below.
```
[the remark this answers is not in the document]
javierhonduco: Good catch haha thanks!

Issue #117, opened 2024-11-25
Add a larger unwind information bucket


javierhonduco:
Some binaries unwind info is too large to fit in the largest bucket. Create a new one to acommodate these.

PR #122, opened 2024-12-01
Add more unwind information buckets

javierhonduco:
Larger buckets to accomodate bigger executables, as well as other bucket sizes to reduce the space waste.

Test Plan
=========

Tried with a large binary, Clickhouse, and it was profiled without issues.


javierhonduco:
Fixes https://github.com/javierhonduco/lightswitch/issues/117

PR #129, opened 2025-01-08
Add unwind info manager implementation and use it

javierhonduco:
* Add unwind info manager implementation
* Use unwind information manager

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -3,7 +3,9 @@ use parking_lot::RwLock;
 use std::collections::hash_map::Entry;
 use std::collections::hash_map::OccupiedEntry;
 use std::collections::HashMap;
+use std::env::temp_dir;
 use std::fs;
+use std::io::ErrorKind;
 use std::mem::size_of;
 use std::mem::ManuallyDrop;
 use std::mem::MaybeUninit;
@@ -20,7 +22,6 @@ use itertools::Itertools;
 use libbpf_rs::num_possible_cpus;
 use libbpf_rs::skel::SkelBuilder;
 use libbpf_rs::skel::{OpenSkel, Skel};
-use libbpf_rs::ErrorKind;
 use libbpf_rs::MapCore;
 use libbpf_rs::MapHandle;
 use libbpf_rs::MapType;
@@ -41,8 +42,8 @@ use crate::process::{
     ProcessStatus,
 };
 use crate::profile::*;
-use crate::unwind_info::compact_unwind_info;
 use crate::unwind_info::log_unwind_info_sections;
+use crate::unwind_info::manager::UnwindInfoManager;
 use crate::unwind_info::types::CompactUnwindRow;
 use crate::util::{get_online_cpus, summarize_address_range};
 use lightswitch_object::{ExecutableId, ObjectFile};
@@ -121,9 +122,11 @@ pub struct Profiler {
     /// evictions which might reduce the quality of the profiles and in more work
     /// for the profiler.
     max_native_unwind_info_size_mb: i32,
+    unwind_info_manager: UnwindInfoManager,
 }
 
 pub struct ProfilerConfig {
+    pub cache_dir: PathBuf,
     pub libbpf_debug: bool,
     pub bpf_logging: bool,
     pub duration: Duration,
@@ -142,7 +145,9 @@ pub struct ProfilerConfig {
 
 impl Default for ProfilerConfig {
     fn default() -> Self {
+        let cache_dir = temp_dir().join("lightswitch");
         Self {
+            cache_dir,
             libbpf_debug: false,
             bpf_logging: false,
             duration: Duration::MAX,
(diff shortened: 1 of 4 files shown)


src/unwind_info/persist.rs
```diff
@@ -118,8 +117,7 @@ pub enum ReaderError {
     Digest,
 }
 
-/// Reads compact information of a bytes slice.
```
javierhonduco, line 121: Oops must have removed it while rebasing, will re-add it

src/unwind_info/persist.rs
```diff
@@ -39,27 +39,26 @@ unsafe impl Plain for Header {}
 /// read path, in case the data is corrupted.
 unsafe impl Plain for CompactUnwindRow {}
 
-/// Writes compact information to a given writer.
```
javierhonduco, line 42: Oops must have removed it while rebasing, will re-add it

src/profiler.rs
```diff
@@ -288,6 +296,11 @@ impl Profiler {
     }
 
     pub fn new(profiler_config: ProfilerConfig, stop_signal_receive: Receiver<()>) -> Self {
+        _ = fs::create_dir(&profiler_config.cache_dir);
```
rdelfin, line 299: Same on this and the one two lines below

src/unwind_info/manager.rs
```diff
@@ -0,0 +1,237 @@
...
+const DEFAULT_MAX_CACHED_FILES: usize = 1_000;
+
+#[derive(Debug, PartialEq, Eq)]
+struct Usage {
+    executable_id: ExecutableId,
+    instant: Instant,
+}
+
+// `BinaryHeap::pop()` returns the biggest element, so reverse it
+// to get the smallest one AKA oldest for both `PartialOrd` and `Ord`.
```
rdelfin, line 25: Doesn't the `Ord` implementation flip it again making this moot?

src/unwind_info/manager.rs
```diff
@@ -0,0 +1,237 @@
...
+/// Provides unwind information with caching on the file system, expiring
+/// older files if there are more than `max_cached_files`.
+pub struct UnwindInfoManager {
+    cache_dir: PathBuf,
+    usage_tracking: BinaryHeap<Usage>,
+    max_cached_files: usize,
+}
+
+impl UnwindInfoManager {
+    pub fn new(cache_dir: &Path, max_cached_files: Option<usize>) -> Self {
```
rdelfin, line 47: Any reason `max_cached_files` only gets set to not-None in tests?

rdelfin (commented):
Thanks for the PR! Just a few questions before approving?

src/profiler.rs
```diff
@@ -142,7 +145,12 @@ pub struct ProfilerConfig {
 
 impl Default for ProfilerConfig {
     fn default() -> Self {
+        let cache_dir = temp_dir().join("lightswitch");
+        debug!("Cache directory {}", cache_dir.display());
+        _ = fs::create_dir(&cache_dir);
```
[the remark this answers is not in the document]
javierhonduco: That's fair, was being lazy because we need to only fail hard / log if the dir doesn't exist already 😄

src/unwind_info/manager.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(other.cmp(self))
+    }
+}
+
+impl Ord for Usage {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        other.instant.cmp(&self.instant)
+    }
+}
```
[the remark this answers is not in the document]
javierhonduco: Mmmm let me see what can I do about this, or at least, add some documentation on this behaviour

src/unwind_info/manager.rs
javierhonduco: `Ord` flips it and `PartialOrd` calls `Ord`'s impl, so there is no double application unless I am missing something 😬

src/unwind_info/manager.rs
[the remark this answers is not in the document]
rdelfin: Yeah, it might be fine as is hoenstly. Just worth confirming

src/unwind_info/manager.rs
rdelfin: @javierhonduco PartialOrd does `other.cmp(self)` and `Ord` also does `other.cmp(self)`. That would flip it right?

rdelfin (approved):

rdelfin:
(discussed offline, happy for it to merge once the comments aboved are addressed

src/unwind_info/manager.rs
javierhonduco: Ugh as discussed offline, I'll fix it, now I see it 🤡

PR #130, opened 2025-01-08
Refactor interface for registering custom label providers

patnebe:
## Context
* The interface for registering custom providers assumes all metadata providers will accept a pid arg. This is not true, especially for providers that return system-level metadata that might be company/deployment/runtime environment specific and not in any way related to the task being profiled.
* To work around this today, one option is for users have to modify `metadata_provider.rs` in their fork to add/invoke logic for custom system metadata providers. This is not ideal.
* A separate interface for registering system-level metadata providers eliminates the need for such code changes. 

## Changes
* Rename `ThreadSafeMetadataProvider` to `ThreadSafeTaskMetadataProvider`
* Introduce a new `SystemMetadataProvider` trait that has a `get_metadata`accepts no arguments and returns a vector of labels.
* Add a new method -- `register_system_metadata_providers()` -- to the `GlobalMetadataProvider` struct.

## Test Plan
* Manual testing + Existing Unit tests.

### DIFF
--- a/lightswitch-metadata/src/metadata_provider.rs
+++ b/lightswitch-metadata/src/metadata_provider.rs
@@ -30,8 +30,7 @@ pub enum TaskMetadataProviderError {
 }
 
 pub trait TaskMetadataProvider {
-    /// Return a vector of labels that apply to
-    ///  the provided task_key.
+    /// Return a vector of labels that apply to the provided task_key.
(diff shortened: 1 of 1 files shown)


javierhonduco (approved):
Makes sense, thanks for the PR!

PR #146, opened 2025-01-27
metadata: Use the thread group id (tgid) for the main task

javierhonduco:
Rather than the process group id (pgrp) used as the main process for signal delivery. While in many instances they match, this code is not correct.

Test Plan
=========
Manual tests


javierhonduco:
cc @patnebe

PR #150, opened 2025-02-04
[Trivial/CHORE] Metadata lib refactor

patnebe:
#### Context
* Refactor the default metadata providers to make their interfaces more consistent with the custom metadata providers

#### Changes
* Move a few shared structs and traits into a `types` mod
* Impl the `*MetadataProvider` traits for the default metadata providers

#### Test Plan
* CI


javierhonduco (approved):
Apologies for the delay, thanks for making the change!

Issue #153, opened 2025-02-06
Make cache directory configurable


javierhonduco:
Cached data is stored under `/tmp` which can be a problem for certain users as it's typically memory-backed and can be quite small which can cause i/o errors. This should be configurable for users that prefer to store the cached data elsewhere.

PR #155, opened 2025-02-06
Make cache directory configurable

javierhonduco:
Cached data is stored under /tmp which can be a problem in certain environments as tmpfs is typically memory-backed and can be quite small which can cause i/o errors. This should be configurable for users that prefer to store the cached data elsewhere.

Also error hard if the directory doesn't exist as it is needed to function properly.

Test Plan
=========

Tested it manually with a different base cache path


javierhonduco:
Fixed by https://github.com/javierhonduco/lightswitch/issues/153

Issue #116, opened 2024-11-25
Improve kernel handling


javierhonduco:
Improve kernel handling, both as the "mappings" it has (the main kernel image and the modules) and from an object file perspective (handling of build ids, etc). This work already started in https://github.com/javierhonduco/lightswitch/tree/kernel-object-improvements, but requires cleaning it up.

javierhonduco:
This is now implemented!

PR #162, opened 2025-02-11
Kernel object improvements

javierhonduco:
See individual commits but the general one

https://blogs.oracle.com/linux/post/whats-inside-a-linux-kernel-core-dump
was invaluable to understand where to read the kernel ASLR offset from.

Test Plan
=========

Checked a couple of symbols on 3 different machines (Ubuntu, RHEL,
Fedora) and the KASLR offset was calculated correctly. Everything worked
on x86 and arm64 too.


javierhonduco:
Fixes https://github.com/javierhonduco/lightswitch/issues/116

src/kernel.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Read and parse the build id of a given kernel module.
+fn _get_module_build_id(module_name: &str) -> Result<BuildId, anyhow::Error> {
+    let mut file = BufReader::new(File::open(format!(
+        "/sys/module/{}/notes/.note.gnu.build-id",
+        module_name
+    ))?);
+    let mut data = Vec::new();
+    file.read_to_end(&mut data)?;
+
+    parse_gnu_build_id_from_notes(&data)
```
rdelfin, line 78: Looking at it, would it make sense to pass the reader directly to the `parse_gnu_build_id_from_notes()`? Or does the iterator not work on readers?

src/kernel.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// Read and parse the build id of the running kernel image.
+/// This can also be done using `perf` with `perf buildid-list --kernel`.
+pub fn kernel_build_id() -> Result<BuildId, anyhow::Error> {
+    let mut file = BufReader::new(File::open("/sys/kernel/notes")?);
+    let mut data = Vec::new();
+    file.read_to_end(&mut data)?;
+
+    parse_gnu_build_id_from_notes(&data)
```
rdelfin, line 97: Same here

rdelfin (approved):
Generally LGTM, just one question

src/kernel.rs
javierhonduco: It doesn't work with readers, only with byte slices:
```Rust
pub fn new(endian: Elf::Endian, align: Elf::Word, data: &'data [u8])
```

The buffers are quite small, and this code doesn't get called often so it should not be an issue performance wise

PR #168, opened 2025-02-27
Add helper to perform fast file identity checks

javierhonduco:
It increases the chances of successful symbolisation for executables in mount namespaces that aren't the root one

Test Plan
=========

CI + ran several manual tests

### DIFF
--- a/src/util/file.rs
+++ b/src/util/file.rs
@@ -1,0 +1,66 @@
+use crate::process::Pid;
+use std::fs;
+use std::io;
+use std::os::unix::fs::MetadataExt;
+use std::path::Path;
+use std::path::PathBuf;
+
+/// Represents a unique file or directory on a Linux machine accross filesystems.
+///
+/// Checking if two files or directories are the same requires byte-for-byte checks
+/// which are expensive in large files. Inodes are unique per device, so read and compare
+/// both as a cheap file identity checks. As they can be re-used (the policy varies across
+/// filesystems) the file size is also checked.
+#[derive(Debug, PartialEq)]
+struct FileId {
+    inode: u64,
+    device: u64,
+    size: u64,
+}
+
+impl FileId {
+    fn new(path: &Path) -> io::Result<Self> {
+        let metadata = fs::metadata(path)?;
+        Ok(FileId {
+            inode: metadata.ino(),
+            device: metadata.dev(),
+            size: metadata.size(),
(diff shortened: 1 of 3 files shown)


src/util/file.rs
```diff
@@ -0,0 +1,61 @@
...
+/// For a given executable path retrieved from procfs /maps, return its absolute path.
+/// If the executable is running in a mount namespace, return the procfs full mount path.
+///
+/// This is useful because mount namespaces can and will go away more often than other mounts
+/// and it can help reduce the chances of race conditions to not rely on the procfs mount path
+/// unless we must.
+pub fn executable_path(pid: Pid, path: &Path) -> io::Result<PathBuf> {
+    // Not using Path join as appending absolute paths will replace the whole path with it, see
+    // https://github.com/rust-lang/rust/issues/16507
+    let procfs_path = format!("/proc/{}/root{}", pid, path.to_string_lossy());
```
rdelfin, line 41: Wonder if it's better to check if it's absolute and remove the `/` in that case so you can safely join. Up to you though, this is definitely an annoying case!

src/util/file.rs
```diff
@@ -0,0 +1,61 @@
...
+    let procfs_path = PathBuf::from(procfs_path);
+    if FileId::new(&procfs_path)? == FileId::new(path)? {
+        return Ok(path.to_path_buf());
+    }
+
+    Ok(procfs_path)
```
rdelfin, line 47: I know I know, extremely nitpick-y

```suggestion
    if FileId::new(&procfs_path)? == FileId::new(path)? {
        Ok(path.to_path_buf())
    } else {
        Ok(procfs_path)
    }
```

rdelfin (approved):
LGTM, just some nits, up to you

src/util/file.rs
javierhonduco: That's a good point! I think it always prefixed by `/` but let me take another look

src/util/file.rs
javierhonduco: I like the previous style better 🫨, but no strong opinions. What are the benefits of this suggestion in your opinion (just curious)

src/util/file.rs
javierhonduco: Confirmed they are always absolute. Adding a runtime check in dev mode though

PR #174, opened 2025-03-05
Implement support for an emergency killswitch

patnebe:
### Context
* Implement an emergency killswitch mechanism to stop the agent if a file is on disk and prevent it from starting 

### Test Plan
* CI
* Manual test


src/cli/killswitch.rs
```diff
@@ -0,0 +1,76 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::{create_dir_all, remove_file, File};
+
+    #[test]
+    fn test_killswitch_enabled() {
+        // Given
+        create_dir_all("/tmp/lightswitch").unwrap();
+        let killswitch = KillSwitch::new(None, /*ignore_killswitch*/ false);
```
javierhonduco, line 36: nit: there's no linter that keeps these comments synced to the source code so they might get out of date. They should not be necessary as most people in the Rust community use [Rust Analyzer](https://rust-analyzer.github.io/) in their editor that shows the argument names in the call sites 
```suggestion
        let killswitch = KillSwitch::new(None, false);
```

src/cli/killswitch.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    #[test]
+    fn test_killswitch_path_override() {
+        // Given
+        let temp_killswitch_file = tempfile::NamedTempFile::new().unwrap();
+        let temp_killswitch_path = temp_killswitch_file.path();
+        let killswitch = KillSwitch::new(
+            Some(temp_killswitch_path.to_str().expect("").to_string()),
+            /*ignore_killswitch*/ false,
```
javierhonduco, line 54: Same a above

src/cli/args.rs
```diff
@@ -146,6 +146,16 @@ pub(crate) struct CliArgs {
...
+    #[arg(
+        long,
+        help = "Override the default path to the killswitch file (/tmp/lighswitch/killswitch) which prevents the profiler from starting"
+    )]
+    pub(crate) killswitch_path_override: Option<String>,
+    #[arg(
+        long,
+        help = "Force the profiler to start even if the system killswitch is enabled"
+    )]
+    pub(crate) unsafe_start: bool,
```
javierhonduco, line 158: Very nice!

src/cli/main.rs
```diff
@@ -119,6 +124,13 @@ fn main() -> Result<(), Box<dyn Error>> {
         std::process::exit(1);
     }
 
+    let killswitch = KillSwitch::new(args.killswitch_path_override, args.unsafe_start);
+    if killswitch.enabled() {
+        info!("Killswitch enabled, will do nothing...");
+        std::thread::sleep(args.duration);
```
javierhonduco, line 130: Why do we need this sleep? I assume it's a safety mechanism in case the process manager doesn't wait in between retries?

src/cli/main.rs
```diff
@@ -171,13 +183,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     ctrlc::set_handler(move || {
         info!("received Ctrl+C, stopping...");
-        let _ = stop_signal_sender.send(());
+        let _ = profiler_stop_signal_sender.send(());
     })
     .expect("Error setting Ctrl-C handler");
 
+    // Start a thread to stop the profiler if the killswitch is enabled
+    let killswitch_ticker = tick(KILLSWITCH_POLL_INTERVAL);
+    thread::spawn(move || loop {
```
javierhonduco, line 195: If we can name this thread so the purpose it's clear that would great (see other thread spawns in the code)

javierhonduco (commented):
Looks good overall! Just left some comments

src/cli/main.rs
patnebe: Yeah, that was the intention. My assumption was that without this, the process will continuously exit and get restarted.

src/cli/killswitch.rs
patnebe: Fair enough. Will update this.

src/cli/main.rs
javierhonduco: I am not sure about this, but don't have strong opinons. One one hand it seems like this should be 100% the responsibility of the init process, but at the same time, having this here to prevent eleveated CPU usage if the config in init is wrong seems sensible... What do you think?

I am happy to merge with the sleep, but could you add for how long it'll sleep in the log message?

javierhonduco:
Thanks so much! Apologies for the delay once again

PR #177, opened 2025-03-10
Register metadata on new process detected

patnebe:
#### Context
* Register task metadata when a new process is first detected

#### Changes
* On new process detected, read all the thread ids for that process and add them to the metadata cache
* Bump `pid_label_cache` size to 5k items as a short-term workaround for a few processes with many threads filling the cache. For example, 50% of the cache is used up by just 17 processes in the below log:
```
2025-03-28T09:38:49.099459Z  INFO lightswitch_metadata::metadata_provider: Task with key TaskKey { pid: 1365, tid: 1420 } registered, labelCacheSize=480, numUniqueProcesses=17
```
* In follow-up PR, I plan to add a different cache for thread-specific metadata to reduce memory used by the labels which are the same for a given process, but duplicated for multiple threads.

#### Test Plan
* Manual Testing


javierhonduco:
This approach makes sense, we can look at different options for the `add_proc` note

tests/integration_test.rs
```diff
@@ -108,7 +111,9 @@ fn test_integration() {
         ..Default::default()
     };
     let (_stop_signal_send, stop_signal_receive) = bounded(1);
-    let mut p = Profiler::new(profiler_config, stop_signal_receive);
+    let metadata_provider: ThreadSafeGlobalMetadataProvider =
```
javierhonduco, line 114: nit: No need to specify the type
```suggestion
    let metadata_provider =
```

javierhonduco (approved):
LGTM, left two comments

src/profiler.rs
```diff
@@ -202,14 +205,6 @@ pub enum AddProcessError {
     ProcfsRace,
 }
 
-impl Default for Profiler {
```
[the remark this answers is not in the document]
patnebe: Done 👍🏽

src/profiler.rs
```diff
@@ -2045,6 +2059,23 @@ impl Profiler {
         };
         self.procs.clone().write().insert(pid, proc_info);
 
+        for thread in proc.tasks().map_err(|_| AddProcessError::ProcfsRace)? {
```
javierhonduco, line 2062: Just realised we might want to change this error handling, but can do this in another PR

patnebe: What do you have in mind for this?

javierhonduco: Even if one of the threads might exit we probably want to register the other ones

PR #179, opened 2025-03-18
Fix off-by-one in BPF pages deletion

javierhonduco:
The last page was not deleted which resulted in the BPF map to eventually be full and no more mappings could be added to it. This commit fixes it by making sure the last page is also deleted, and adding integration tests to verify that all the necessary BPF maps are cleaned up.

Test Plan
=========

Added tests and ran for a while whilst spawning many new processes with different unwind information without issues.

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -9,6 +9,7 @@ use std::env::temp_dir;
 use std::fs;
 use std::fs::File;
 use std::io::ErrorKind;
+use std::iter;
 use std::mem::size_of;
 use std::mem::ManuallyDrop;
 use std::mem::MaybeUninit;
@@ -760,8 +761,10 @@ impl Profiler {
                 debug!("marking process {} as exited", pid);
                 proc_info.status = ProcessStatus::Exited;
 
-                // Delete process, todo track errors.
-                let _ = Self::delete_bpf_process(&self.native_unwinder, pid);
+                let err = Self::delete_bpf_process(&self.native_unwinder, pid);
+                if let Err(e) = err {
+                    debug!("could not remove bpf process due to {:?}", e);
+                }
 
                 for mapping in &mut proc_info.mappings.0 {
                     let mut object_files = self.object_files.write();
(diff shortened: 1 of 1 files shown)


src/profiler.rs
```diff
@@ -1995,3 +2004,114 @@ impl Profiler {
...
+        assert_eq!(profiler.native_unwinder.maps.outer_map_6.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_7.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_8.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_9.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_10.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_11.keys().count(), 0);
+        assert_eq!(profiler.native_unwinder.maps.outer_map_12.keys().count(), 0);
+
+        assert_eq!(
+            profiler.native_unwinder.maps.exec_mappings.keys().count(),
```
javierhonduco, line 2104: This fails on arm64, very sus https://github.com/javierhonduco/lightswitch/actions/runs/13923652981/job/38962536833?pr=179 cc @rdelfin will debug later as I don't have much time this week

javierhonduco: Luckily this is not seem as dodgy as I feared, on arm64 we are not cleaning vDSO, which doesn't contain unwind info. This is due to subpar special condition handling (no unwind info should mean no mappings are written)

```
(nix:nix-shell-env) javierhonduco@lima-debian:/Users/javierhonduco/src/lightswitch$ sudo target/debug/lightswitch show-unwind /tmp/lightswitch/dumped-vdso
thread 'main' panicked at src/cli/main.rs:298:49:
called `Result::unwrap()` on an `Err` value: no eh_frame section found
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
``

PR #184, opened 2025-03-19
Improve error handling dealing with unwind info

javierhonduco:
This was discovered thanks to
https://github.com/javierhonduco/lightswitch/commit/37a0dc8de23214b9573a1ff5fd3c1bea13043672. Basically the current poor error handling was causing data in BPF maps that should be deleted to be left behind. This surfaced on arm64 as vDSO sections don't have unwind information and this error code path ran.

Test Plan
=========

The test now passes on arm64.

### DIFF
--- a/src/bpf/profiler.bpf.c
+++ b/src/bpf/profiler.bpf.c
@@ -192,7 +192,7 @@ find_page(mapping_t *mapping, u64 object_relative_pc, u64 *low_index, u64 *high_
     }
   }
 
-  LOG("[error] could not find page");
+  LOG("[error] could not find page for executable_id: %llx at file_offset: %llx", mapping->executable_id, object_relative_pc);
   bump_unwind_error_chunk_not_found();
   return NULL;
 }
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -49,7 +49,6 @@ use crate::process::{
     ProcessStatus,
 };
 use crate::profile::*;
-use crate::unwind_info::log_unwind_info_sections;
 use crate::unwind_info::manager::UnwindInfoManager;
 use crate::unwind_info::types::CompactUnwindRow;
 use crate::util::executable_path;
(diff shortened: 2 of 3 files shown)


src/bpf/profiler.bpf.c
```diff
@@ -192,7 +192,7 @@ find_page(mapping_t *mapping, u64 object_relative_pc, u64 *low_index, u64 *high_
     }
   }
 
-  LOG("[error] could not find page");
+  LOG("[error] could not find page for executable_id: %llx at file_offset: %llx", mapping->executable_id, object_relative_pc);
```
javierhonduco, line 195: Aid during debugging that is worth keeping

src/profiler.rs
```diff
@@ -1438,7 +1458,17 @@ impl Profiler {
...
-            self.add_unwind_information_for_executable(mapping.executable_id);
+            if let Err(e) = self.add_unwind_information_for_executable(mapping.executable_id) {
+                if e == AddUnwindInformationError::NoUnwindInfoKnownNaughty {
+                    return;
+                }
+                warn!(
+                    "error adding unwind information for executable {:x} due to {:?}",
+                    mapping.executable_id, e
+                );
+                // TODO: maybe cleanup written maps.
```
javierhonduco, line 1469: I need to make up my mind about this... cleaning up everything means no profiling at all is possible for this process, but not cleaning up means that there will be errors that can only be explained with a deeper look at logs. We might be able to change the latter if we account for errors dealing with mappings in the process info

src/profiler.rs
```diff
@@ -1780,7 +1796,7 @@ impl Profiler {
                     let object_file = match ObjectFile::new(&exe_path.clone()) {
                         Ok(f) => f,
                         Err(e) => {
-                            warn!("object_file {} failed with {}", exe_path.display(), e);
```
javierhonduco, line 1783: Race conditions happen often here and it's quite chatty

src/unwind_info/types.rs
```diff
@@ -20,7 +20,7 @@ pub enum CfaType {
 #[derive(Debug, Default, Copy, Clone, PartialEq)]
 pub enum RbpType {
     #[default]
-    Unknown = 0,
+    Unchanged = 0,
```
javierhonduco, line 23: The name was wrong (see https://github.com/javierhonduco/lightswitch/blob/92543b43209398b250396291f2e9b646c98bf599/src/bpf/profiler.h#L69)

src/profiler.rs
```diff
@@ -1458,18 +1488,12 @@ impl Profiler {
...
-                "unwind info already loaded for executable id: {:x}",
-                executable_id
-            );
-            return;
-        } else {
-            debug!(
-                "unwind info not loaded for executable id: {:x}",
-                executable_id
-            );
+            return Ok(AddUnwindInformationResult::AlreadyLoaded);
```
rdelfin, line 1496: I don't see the debug you removed added back anywhere. Was that intentional?

rdelfin (approved):
Fantastic improvement!

src/profiler.rs
javierhonduco: Yes, moved this to the call-site

PR #193, opened 2025-03-26
Add optional bearer token authentication for write requests

javierhonduco:
Add optional bearer token authentication for write requests to the backend

Test Plan
=========

CI


rdelfin (approved):
Very demure, very mindful

PR #203, opened 2025-04-03
Add support for Go

javierhonduco:
See commits for this PR, but the summary is:

- added logic to differentiate between runtimes and add unwind
  information
- enabled and pinned our symbolization library
- adapted `to_pages` to support large gaps
- fixed address normalization

Test Plan
=========

CI, will add Go integration tests later on. I've ran manual tests
with the test application as well as tailscaled without any issues.


javierhonduco:
Proof that it works 😄  https://github.com/tailscale/tailscale/issues/15492

Issue #205, opened 2025-04-04
Use `starts_with` rather than `contains` to find the stop frames for Go


javierhonduco:
Forgot to add this minor performance optimisation

javierhonduco:
Oops reopened it by accident

PR #208, opened 2025-04-04
Detect stop frames faster

javierhonduco:
Test Plan
=========

Go profiling works


javierhonduco:
Fixes https://github.com/javierhonduco/lightswitch/issues/205

PR #209, opened 2025-04-04
Use `mmap()` to write unwind info and make it 70x faster

javierhonduco:
This commit changes how unwind information is written to the BPF maps where it's stored. Instead of using batched updates it uses the `BPF_F_MMAPABLE` option introduced in kernel version 5.5 [0], which was released ~5y ago, so it's supported by all the kernels we care about.

Saying that the improvement is "good" would be a significant understatement. It's now 70x faster! Batched BPF updates seem to be way slower than I expected. I couldn't believe the CPU profiles obtained with lightswitch itself, so it wrote some benchmarks and also ran `strace` and `perf trace` to make sure that the numbers made sense.

Note that thanks to using `mmap()` reduces the amount of data buffering. Before this commit, we had to read from disk -> chunks -> bpf maps. Now the intermediate buffer is no longer needed.

Potentially using larger mmap writes would be faster due to `memcpy` having optimizations for large memory copies, but this is so fast right not compared to the previous implementation that I don't think this would be necessary.

The only downside I can think of is that now the allocations that back the BPF arrays need to be page aligned, and in case of larger pages, beyond 4KB there could be some wastage. For the standard page size, this is not really an issue as unwind information can get quite large (dozens of MBs is not unheard of).

Test Plan
=========

## Benchmark comparison

image

## Batched update with single batch to minimise syscall overhead

image

## mmaped version

image

- [0]: https://github.com/torvalds/linux/commit/fc9702273e2edb90400a34b3be76f7b08fa3344b

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1322,6 +1322,7 @@ dependencies = [
  "glob",
  "libbpf-cargo",
  "libbpf-rs",
+ "libbpf-sys",
  "libc",
  "nix",
  "perf-event-open-sys",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ nix = { version = "0.29.0" }
 libbpf-cargo = { version = "0.25.0-beta.1" }
 glob = "0.3.2"
 ring = "0.17.14"
+libbpf-sys = "1.5.0"
 
 [dependencies]
 gimli = "0.31.1"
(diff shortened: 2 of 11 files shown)


javierhonduco:
Wanted to understand why batched updates had these performance characteristics:

image

Seems like most of the time it's spent copying data (which is not terribly surprising). If we look at the line of the code responsible for copying from user to kernel:

https://github.com/torvalds/linux/blob/38fec10eb60d687e30c8c6b5420d86e8149f7557/kernel/bpf/syscall.c#L1958

```c
[...]
		if (copy_from_user(key, keys + cp * map->key_size,
		    map->key_size) ||
		    copy_from_user(value, values + cp * value_size, value_size))
			break;
[...]
```

Seems like the copying is done a single element at a time and `copy_from_user` is called for each. I am guessing this was not optimised for BPF arrays to not have to allocate more memory in the kernel and perhaps increase the opportunities of running `cond_resched()` and RCU synchronisation? But either way it seems like perhaps this could be batched in the kernel, too? That would seem like the potential tradeoff here.

anakryiko:
> cc @anakryiko, as you implemented mmap-able BPF arrays and this win is too good to keep it to myself, thanks! :)

yeah, doing syscall-based batch updates from user space with BPF_MAP_TYPE_ARRAY doesn't make much sense, always go mmap route, if you can.

Issue #204, opened 2025-04-03
Adapt task_pt_regs for FRED


javierhonduco:
https://tanelpoder.com/posts/ebpf-pt-regs-error-on-linux-blame-fred and also use `bpf_task_pt_regs()` if available

PR #210, opened 2025-04-05
Use `bpf_task_pt_regs` if available

javierhonduco:
In kernel context (syscalls, interrupts, etc) the unwinder has to do additional work to fish out the userspace registers. We do this by replicating the logic in the kernel by hand. This works fine, but unfortunately there's been some ABI changes [0]. Rather than adjusting it and have to keep it in sync, use the `bpf_task_pt_regs` helper to do this.

It was added in 2021 and the ABI changes are more recent, so keeping the current logic when the helper is not available as is.

- [0]: https://github.com/javierhonduco/lightswitch/issues/204


javierhonduco:
Fixes https://github.com/javierhonduco/lightswitch/issues/204

PR #216, opened 2025-04-10
Improve rate-limiting efficacy of new code regions events

javierhonduco:
See code comments, but fewer events to process due to more effective rate-limiting without any correctness tradeoffs.

Test Plan
=========

Verified that the mask is correct + CI


patnebe (approved):
Looks good :shipit:

PR #217, opened 2025-04-10
Improve error handling for critical BPF map ops

javierhonduco:
This is not deal with right now (cleaning half-written data, etc) but we should definitely never panic in these code paths

Test Plan
=========

CI


patnebe (approved):
Intent sounds reasonable to me.

PR #236, opened 2025-05-01
ci: Show nix version

javierhonduco:
This can aid in troubleshooting issues arising from running a different nix version


javierhonduco:
cc @patnebe

patnebe (approved):
Thanks!

PR #241, opened 2025-05-02
Refactor stack + key into new struct

patnebe:
#### Context
* Create a simple struct containing the stack info. This is used in https://github.com/javierhonduco/lightswitch/pull/245 which moves the sample aggregation out of bpf to userspace. This struct is the raw sample payload sent from bpf to userspace.

#### Changes
* Refactor the relevant subset of the `unwind_state_t` into a `raw_stack_t` struct.

#### Test Plan
* Ran locally to smoke test that nothing obvious is broken 
* CI + existing tests.

### DIFF
--- a/src/aggregator.rs
+++ b/src/aggregator.rs
@@ -14,7 +14,7 @@ impl Aggregator {
             return Vec::new();
         }
 
-        let mut sample_hash_to_aggregated = HashMap::new();
+        let mut sample_hash_to_aggregated: HashMap<u64, RawAggregatedSample> = HashMap::new();
         for sample in raw_samples {
             if sample.ustack.is_none() & sample.kstack.is_none() {
                 warn!(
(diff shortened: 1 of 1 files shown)


src/bpf/profiler.bpf.c
```diff
@@ -340,7 +340,7 @@ bpf_map_lookup_or_try_init(void *map, const void *key, const void *init) {
 static __always_inline void add_stack(struct bpf_perf_event_data *ctx,
                                       u64 pid_tgid,
                                       unwind_state_t *unwind_state) {
-  stack_count_key_t *stack_key = &unwind_state->stack_key;
+  stack_count_key_t *stack_key = &unwind_state->raw_stack.stack_key;
```
javierhonduco, line 343: nit: at this stage there's no raw stack and processed ones, it's just a stack :D 
```suggestion
  stack_count_key_t *stack_key = &unwind_state->stack.stack_key;
```

patnebe: Makes sense. Tried renaming this struct from `raw_stack_t` to `stack_t` and there was typedef redifinition error

```
./src/bpf/profiler.h:182:3: error: typedef redefinition with different types ('struct stack_t' vs 'struct sigaltstack')
         } stack_t;
           ^
         ./src/bpf/vmlinux_x86.h:17532:28: note: previous definition is here
         typedef struct sigaltstack stack_t;
```

Maybe we can rename the typedef to `stack_sample_t` (open to better naming suggestions)? Or do we keep the typedef as `raw_stack_t` and rename just the member variable in the unwind_state?

javierhonduco: Either works, don't want to block you here, we can always change it down the line!

patnebe: Thanks!

PR #242, opened 2025-05-02
Add uspace aggregation component

patnebe:
#### Context
* Logic for userspace aggregation. This is used in https://github.com/javierhonduco/lightswitch/pull/245.

#### Changes
* Create a new component to handle the aggregation of samples in user space

#### Test Plan
* Unit tests.
* CI + existing tests.

### DIFF
--- a/src/aggregator.rs
+++ b/src/aggregator.rs
@@ -1,0 +1,289 @@
+use std::hash::DefaultHasher;
+use std::{collections::HashMap, hash::Hash, hash::Hasher};
+
+use tracing::warn;
+
+use crate::profile::{RawAggregatedProfile, RawAggregatedSample, RawSample};
+
+#[derive(Default)]
+pub struct Aggregator {}
+
+impl Aggregator {
+    pub fn aggregate(&self, raw_samples: Vec<RawSample>) -> RawAggregatedProfile {
+        if raw_samples.is_empty() {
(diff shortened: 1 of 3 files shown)


src/aggregator.rs
```diff
@@ -0,0 +1,289 @@
...
+#[derive(Default)]
+pub struct Aggregator {}
+
+impl Aggregator {
+    pub fn aggregate(&self, raw_samples: Vec<RawSample>) -> RawAggregatedProfile {
+        if raw_samples.is_empty() {
+            return Vec::new();
+        }
+
+        let mut sample_hash_to_aggregated: HashMap<u64, RawAggregatedSample> = HashMap::new();
```
javierhonduco, line 17: nit: No need to add the type if it can be inferred

patnebe: Thanks. Will address in a follow up.

patnebe: Looks like this is needed. Removing it results in a compilation error for the closure below

```rust
error[E0282]: type annotations needed for `&mut _`
  --> src/aggregator.rs:33:30
   |
33 |                 .and_modify(|aggregated_sample| aggregated_sample.count += 1)
   |                              ^^^^^^^^^^^^^^^^^  ----------------------- type must be known at this point
```

PR #267, opened 2025-06-30
sym: Fetch filename and line number if available

javierhonduco:
Certain symbolization sources such as DWARF can provide us with filename and line numbers but we weren't doing anything with this data. This commit stores it in a new added structure and passes it to the pprof builder. The flamegraph in svg format isn't changed and will only show the function names

Test Plan
=========

existing tests and manually checked with the `pprof` tool that stacks can be aggregated using the filenames and line numbers


rdelfin (approved):
Mostly LGTM, just a couple things

javierhonduco:
Will merge but happy to re-evaluate this

Issue #266, opened 2025-06-28
Add support for unwind expressions of type `cfa = *($rsp + offset) + value`


javierhonduco:
These are quite common in lightswitch itself and other workloads, and currently they are the top 1 reason of unwinding errors

Issue #264, opened 2025-06-28
"Performance Profile Visualization Challenge" tracking issue, including improved Zig support


javierhonduco:
Bumped into a very interesting "call for profiling" for a project that uses Zig (https://matklad.github.io/2025/05/06/performance-profile-visualization-challenge.html). Profiling it didn't work. The following should address the issues preventing profiling [Tigerbeetle](https://tigerbeetle.com/) (and many other workloads!)

- [x] Fix the load address calculation which is wrong for this workload. This was always an approximation but worked fine for the vast majority of executables I cared about. It was never fully correct so it's time to fix this (https://github.com/javierhonduco/lightswitch/pull/272).
- [x] Older versions of Zig don't mark the first frame as "stop unwinding", address this (https://github.com/javierhonduco/lightswitch/pull/269).
- [x] The line number and path are never written to the pprof file. I somehow never needed line-level information, but clearly this can be very useful (https://github.com/javierhonduco/lightswitch/pull/267).
- [x] Add an option for the flamegraph to show paths and file numbers (https://github.com/javierhonduco/lightswitch/pull/274).

javierhonduco:
Now that all the pieces of the puzzle are in place:

```bash
[javierhonduco@fedora tigerbeetle]$ git show
commit 0e8be09fb44058c4b37af1fa66458f692e840f8b (HEAD)
Merge: 0c633beb2 34caad945
Author: Alex Kladov <aleksey.kladov@gmail.com>
Date:   Mon May 5 14:53:51 2025 +0000

    Merge pull request #2926 from tigerbeetle/matklad/git-review-split
    
    git-review: add split subcommand
```

```bash
[javierhonduco@fedora tigerbeetle]$ /usr/bin/time ./zig/zig build test -- Cluster
[5/7] steps
└─ [1/52] run test
   └─ vsr.replica_test.test.Cluster: recovery:
```

```bash
(nix:nix-shell-env) [javierhonduco@fedora lightswitch]$ cargo run  -- --pids=$(pidof test) --flamegraph-aggregation=all
```

[...]

PR #269, opened 2025-06-30
Fully support Zig

javierhonduco:
While profiling a Zig workload
(https://matklad.github.io/2025/05/06/performance-profile-visualization-challenge.html) unwinding catastrophically fails, this is due to a couple of issues documented on https://github.com/javierhonduco/lightswitch/issues/264. This commit focuses on adding support for Zig executables compiled with the Zig compiler before commit [130f7c2ed8e3358e24bb2fc7cca57f7a6f1f85c3](https://github.com/ziglang/zig/commit/130f7c2ed8e3358e24bb2fc7cca57f7a6f1f85c3) as the first function, `start` did not have the right DWARF unwind information.

In order to remediate this on our side, we attempt to recognise Zig binaries with simple symbol-based heuristics, as there is no metadata of any kind in the ELF files that we could use to know whether it was compiled with the Zig compiler and if the version used contains this bug.

The address range for the first function is read and the right unwind information is then overriden for this file offset. We can't tell if the compiler used has the bug or not until we aren't processing the unwind information and given that all this is cached, the additional work of always having traversing the symbol table is not a problem. I've verified that this is the case with some self-profiles.

Notes
=====

- This is not enough to be able to successfully profile Tigerbeetle, more PRs coming, see issue and "Test Plan" there;
- The persisted unwind information for Zig should be reset. Will update the version in another PR as https://github.com/javierhonduco/lightswitch/issues/266 needs a version bump too

### DIFF
--- a/lightswitch-object/src/object.rs
+++ b/lightswitch-object/src/object.rs
@@ -32,6 +32,14 @@ pub struct ElfLoad {
 pub enum Runtime {
     /// C, C++, Rust, Fortran.
     CLike,
+    /// Zig. Needs special handling because before [0] the top level frame (`start`) didn't have the
+    /// right unwind information
+    ///
+    /// [0]: https://github.com/ziglang/zig/commit/130f7c2ed8e3358e24bb2fc7cca57f7a6f1f85c3
+    Zig {
+        start_low_address: u64,
+        start_high_address: u64,
+    },
     /// Golang.
     Go(Vec<StopUnwindingFrames>),
 }
@@ -125,6 +133,29 @@ impl ObjectFile {
         if self.is_go() {
             Runtime::Go(self.go_stop_unwinding_frames())
         } else {
+            let mut is_zig = false;
+            let mut zig_first_frame = None;
+
+            for symbol in self.object.symbols() {
+                let Ok(name) = symbol.name() else { continue };
+                if name.starts_with("__zig") {
+                    is_zig = true;
+                }
+                if name == "_start" {
+                    zig_first_frame = Some((symbol.address(), symbol.address() + symbol.size()));
+                }
+
+                // Once we've found both Zig markers we are done. Not that this is a heuristic and it's
+                // possible that a Zig library is linked against code written in a C-like language. In this
+                // case we might be rewriting unwind information that's correct. This won't have a negative
+                // effect as `_start` is always the first function.
+                if is_zig && let Some((low_address, high_address)) = zig_first_frame {
+                    return Runtime::Zig {
+                        start_low_address: low_address,
+                        start_high_address: high_address,
+                    };
+                }
+            }
             Runtime::CLike
         }
     }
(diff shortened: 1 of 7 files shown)


lightswitch-object/src/object.rs
```diff
@@ -125,6 +133,26 @@ impl ObjectFile {
         if self.is_go() {
             Runtime::Go(self.go_stop_unwinding_frames())
         } else {
+            let mut is_zig = false;
+            let mut zig_first_frame = None;
+
+            for symbol in self.object.symbols() {
+                let Ok(name) = symbol.name() else { continue };
+                if name.starts_with("__zig") {
```
rdelfin, line 141: Strange thought, but could this happen if you link zig in as a library? Can you easily do that?

src/unwind_info/convert.rs
```diff
@@ -219,54 +222,57 @@ impl<'a> CompactUnwindInfoBuilder<'a> {
                     _ => continue,
                 }
 
+                if let Some(first_frame_override) = self.first_frame_override {
+                    if compact_row.pc == first_frame_override.0 {
+                        println!("func needs stuff");
```
rdelfin, line 227: Should this be here? Or can you make it something like a `log` message instead?

src/unwind_info/manager.rs
```diff
@@ -245,8 +254,11 @@ mod tests {
         file.write_all(&[0; 20]).unwrap();
 
         // Make sure the corrupted one gets replaced and things work.
-        let manager_unwind_info =
-            manager.fetch_unwind_info(&PathBuf::from("/proc/self/exe"), ExecutableId(0xFABADA));
+        let manager_unwind_info = manager.fetch_unwind_info(
+            &PathBuf::from("/proc/self/exe"),
+            ExecutableId(0xFABADA),
```
rdelfin, line 259: Jajajaj fabada

src/unwind_info/persist.rs
```diff
@@ -74,9 +76,16 @@ impl Writer {
         Ok(unwind_info)
     }
 
-    fn read_unwind_info(&self) -> Result<Vec<CompactUnwindRow>, WriterError> {
-        compact_unwind_info(&self.executable_path.to_string_lossy())
-            .map_err(|e| WriterError::UnwindInfoGeneric(e.to_string()))
+    // bad name!
```
rdelfin, line 79: It is?

rdelfin (commented):
Just a few questions before approving

lightswitch-object/src/object.rs
```diff
@@ -125,6 +133,26 @@ impl ObjectFile {
         if self.is_go() {
             Runtime::Go(self.go_stop_unwinding_frames())
         } else {
+            let mut is_zig = false;
+            let mut zig_first_frame = None;
+
+            for symbol in self.object.symbols() {
```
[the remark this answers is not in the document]
rdelfin: Yeah, that's a fair decision!

lightswitch-object/src/object.rs
rdelfin: Yeah, might be worth dropping a comment mentioning this caveat. As long as the fallback works, that makes sense.

PR #273, opened 2025-07-01
Add support for `*($rsp + offset) + value` DWARF expressions

javierhonduco:
These are hit quite regularly on executables such as lightswitch itself and others I've tested. Trying to get successful unwindings close to 100% for the workloads running on my box.

Test Plan
=========

Manually tested that we can unwind at PCs that have that expression for their CFA.


javierhonduco:
Closed https://github.com/javierhonduco/lightswitch/issues/266

PR #294, opened 2025-08-01
Improve perf fd variable name

javierhonduco:
Test Plan
=========

CI

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -2064,21 +2064,21 @@ impl Profiler {
     }
 
     pub fn setup_perf_events(&mut self) {
-        let mut prog_fds = Vec::new();
+        let mut perf_fds = Vec::new();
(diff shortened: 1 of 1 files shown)


src/profiler.rs
```diff
@@ -2064,21 +2064,21 @@ impl Profiler {
     }
 
     pub fn setup_perf_events(&mut self) {
-        let mut prog_fds = Vec::new();
+        let mut perf_fds = Vec::new();
```
rdelfin, line 2067: terrible name, 0/10

gmarler: Perhaps `perf_event_fds` instead?

rdelfin: I'm kidding :smile: 
Looks clear enough

javierhonduco: The previous name was indeed terrible!

PR #316, opened 2025-10-07
Add clang-format to prevent breaking eBPF format changes

gmarler:
Depending on whether your editor uses `clang-format` or not to reformat certain code files (mine does), if you don't have a sensible configuration for `clang-format`, you might end up with breaking changes in your eBPF code - like reordering includes.

This PR adds sensible defaults, particularly `SortIncludes: 'false'`, although @javierhonduco may want to make changes to indent widths and the like.

I've tested this against most of the BPF files, and it has prevented breaking changes, as well as kept most of the formatting the way it is.


javierhonduco:
This is a good idea, a C formatter is long overdue. That being said I am not sure about this configuration. In the past I've used this config (https://github.com/javierhonduco/rbperf/blob/main/.clang-format). I am open to changes to it, but I am quite partial to the Google style.

As part of these changes, we should also ensure there's an easy way to re-format the files via the CLI (see https://github.com/javierhonduco/rbperf/blob/c0d0956fdeebe5c8280da991b2a792f32249e721/justfile#L31) as well as checking for style in CI.

Would this be something you would want to do?

gmarler:
Finally fixed this:
- Added `just` to the flake (and no formatting changes)
- Added the `justfile`
- Only reformatted the contents of `src/bpf/`

Probably have to rebase from main, since there have been dependabot changes since I created this branch - just let me know what commit to use from main, if needed.

src/bpf/profiler.h
```diff
@@ -38,71 +38,69 @@ _Static_assert(MAX_BINARY_SEARCH_DEPTH >= UNWIND_INFO_PAGE_BIT_LEN, "unwind tabl
...
-#define CFA_TYPE_RSP                    2
-#define CFA_TYPE_CFA_TYPE_UNSUP_EXP     3
-#define CFA_TYPE_PLT1                   4
-#define CFA_TYPE_PLT2                   5
-#define CFA_TYPE_DEREF_AND_ADD          6
-#define CFA_TYPE_END_OF_FDE_MARKER      7
-#define CFA_TYPE_UNSUP_REGISTER_OFFSET  8   // not used in the unwinder yet.
-#define CFA_TYPE_OFFSET_DID_NOT_FIT     9
+#define CFA_TYPE_RBP 1
+#define CFA_TYPE_RSP 2
```
javierhonduco: Sorry just noticed this, could you add `AlignConsecutiveMacros: true`? Thanks!

Issue #323, opened 2025-10-10
munmap tracer programs cause panic on RHEL 9.x


gmarler:
lightswitch v0.3.1, when started on RHEL 9.6, immediately panics with the following (with RUST_BACKTRACE=full):

```
...
2025-10-10T11:50:53.422374Z DEBUG lightswitch::profiler: adding kernel module [vmlinux] [0xffffffffa0600000 - 0xffffffffa1600000)

thread 'main' panicked at src/profiler.rs:721:31:
attach tracers: Error: Permission denied (os error 13)
stack backtrace:
   0:     0x555555c59992 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h4af010e4b1354d4d
   1:     0x555555c7f7b3 - core::fmt::write::h17347e42c9203d6d
   2:     0x555555c55b63 - std::io::Write::write_fmt::h05610498df74ae59
   3:     0x555555c597e2 - std::sys::backtrace::BacktraceLock::print::h0286e2cc1356b990
   4:     0x555555c5b2a9 - std::panicking::default_hook::{{closure}}::hb566dc96699e42f6
   5:     0x555555c5b08b - std::panicking::default_hook::h1b340523e54d7d17
   6:     0x5555557d4421 - lightswitch::panic_thread_hook::{{closure}}::hca7a44bbcd05f0af
   7:     0x555555c5bd7e - std::panicking::rust_panic_with_hook::hceef4321c6f4ad8a
   8:     0x555555c5ba4a - std::panicking::begin_panic_handler::{{closure}}::hf9fd67a226c3bb3d
   9:     0x555555c59e89 - std::sys::backtrace::__rust_end_short_backtrace::h52410ec1fdc70787
  10:     0x555555c5b6dd - __rustc[507298ddafb0ef2f]::rust_begin_unwind
  11:     0x555555c7d1d0 - core::panicking::panic_fmt::hf04b323265684a46
  12:     0x555555c7d596 - core::result::unwrap_failed::hdf92484becbba54e
  13:     0x555555806bd4 - lightswitch::profiler::Profiler::run::h8ede74104368aa90
  14:     0x5555557d62ab - lightswitch::main::he67577b2f10368df
  15:     0x5555557db546 - std::sys::backtrace::__rust_begin_short_backtrace::h852a5d5477f6819b
[...]

PR #327, opened 2025-10-14
Fix issues with munmap() eBPF tracer programs

gmarler:
First fix - ensure `struct munmap_entry_args` are CO-RE compatible across all kernels.


javierhonduco:
Closing since as per we discussed, this bug was introduced in RHEL 9 and it's most likely due to a bug in this distribution's verifier. More context in https://github.com/libbpf/libbpf-rs/issues/1259. Could you add the Red Hat issue tracker information once you have it?

I am not against having a way to deal with this RHEL 9 bug if we must, but let's discuss other approaches in [the issue](https://github.com/javierhonduco/lightswitch/issues/323)

PR #341, opened 2025-11-28
Add from_bytes for exec_mappings_key so we can extract PID

gmarler:
A conversion `from_bytes` of a exec_mappings_key back into the Rust struct is needed for cases when we need to check whether a given PID still has entries in the `exec_mappings` BPF map.


gmarler:
Wanted to use `anyhow::Result()` here, but it's complaining that there's no conversion between `Plain::Error` and `std::error::Error`.

PR #345, opened 2025-12-02
Ensure logic of maybe_evict_process is as intended

gmarler:
Not clear if variable `should_evict` should be set to true if `too_many_processes` is set to false.

This might indeed be the intent, but don't understand the reasoning behind it.

Also changed `if_too_many_procs` to just `too_many_procs`, because `if if_too_many_procs` looks really weird.

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -1745,17 +1745,20 @@ impl Profiler {
         }
     }
 
-    /// Evicts a process. If *if_too_many_procs* is true, this will only be done if there are more
+    /// Evicts a process. If *too_many_procs* is true, this will only be done if there are more
     /// processes with  [`ProcessStatus::Running`] status than the maximum number of processes, [`MAX_PROCESSES`].
     /// Returns false only if an eviction is necessary but not enough time has elapsed since the last one.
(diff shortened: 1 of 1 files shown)


src/profiler.rs
```diff
@@ -1745,17 +1745,20 @@ impl Profiler {
...
         let running_procs = procs
             .iter()
             .filter(|e| e.1.status == ProcessStatus::Running);
-        let should_evict = if if_too_many_procs {
+        let should_evict = if too_many_procs {
+            // Only evict if we *still* have too many processes - otherwise don't
             running_procs.clone().count() >= MAX_PROCESSES as usize
         } else {
+            // Evict even if we don't know if there are too many processes?
+            // Or should this be false?
```
javierhonduco: Yes, that's correct, this would be to call eviction even if there are not too many procs, see the callsite:

```rust
            debug!("eviction result {}", self.maybe_evict_process(false));
```

gmarler: Yes, the logic is sound - the only thing left is to deal with the predicate name, so we don't end up with `if if_...` - I'll update the comments about the intent.

If we can come up with a new predicate name, that'll be the only thing added as part of this PR aside from helpful additional comments.

PR #350, opened 2025-12-11
Process and Mappings Cleanup with Resource Reporting

gmarler:


src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
         start.elapsed()
     }
 
+    fn report_resource_consumption(&mut self) {
+        let (mut exited_procs, mut running_procs) = (0, 0);
+        let procs_guard = self.procs.read();
+        for (_pid, proc_info) in procs_guard.iter() {
```
javierhonduco: nit
```suggestion
        for proc_info in procs_guard.values() {
```

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+            if proc_info.status == ProcessStatus::Exited {
```
javierhonduco: nit: Use match to ensure we cover all variants

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+        );
+        info!(
+            "{} new processes detected this session, {} new processes detected since profiler startup",
+            self.new_proc_per_session, self.new_proc_total
+        );
+        info!(
+            "{} process exits this session; {} process exits since profiler startup",
+            self.exit_proc_per_session, self.exit_proc_total
+        );
+        info!(
```
javierhonduco: IMO this should be a `debug` log

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+            "exec_mappings usage: {}/{}",
+            self.native_unwinder.maps.exec_mappings.keys().count(),
+            self.native_unwinder
+                .maps
+                .exec_mappings
+                .info()
+                .unwrap()
+                .info
+                .max_entries
```
javierhonduco: This number is 'static' once the BPF map is created, right?

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+            self.exit_proc_per_session, self.exit_proc_total
+        );
+        info!(
+            "exec_mappings usage: {}/{}",
+            self.native_unwinder.maps.exec_mappings.keys().count(),
+            self.native_unwinder
+                .maps
+                .exec_mappings
+                .info()
+                .unwrap()
```
javierhonduco: If possible, let's avoid adding new unwraps to the code. I am aware the codebase is littered with some uncessary unwraps which I am planning to tackle soon. If you think it's an invariant that should always hold true, `expect("$invariant")` is preferred in this project

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+    }
+
     pub fn handle_process_exit(&mut self, pid: Pid, partial_write: bool) {
         // TODO: remove ratelimits for this process.
+        // This handler can be called before we've had a chance to register the pid in the
+        // first place, so we should just put the PID in the deletion_scheduler, and do any
+        // work after a couple of sessions have elapsed.
+        self.deletion_scheduler
+            .write()
+            .add(ToDelete::Process(Instant::now(), pid, partial_write));
```
javierhonduco: Is the timestamp correct? Need to check the API because I don't remember atm, but leaving a note so I don't forget

src/profiler.rs
```diff
@@ -1712,8 +1785,15 @@ impl Profiler {
...
                 self.add_unwind_info_for_process(pid);
             }
-            Err(_e) => {
-                // probabaly a procfs race
+            Err(e) => {
+                // probably a procfs race
+                // Could be other things too, like eviction failures
+                match e {
+                    AddProcessError::ProcfsRace => (),
+                    _ => {
```
javierhonduco: style nit: prefer `if let` to deconstruct this enum if we want to only discrimiate against one variant (or in this case, everything but one variant)

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
...
+        // Pop off any processes that we've kept around long enough after they've exited
+        // Where "long enough" is 2 sessions worth
+        let pending_duration = self.session_duration.checked_mul(2).unwrap();
+        let mut pending_deletion: Vec<(Pid, bool)> = vec![];
+        loop {
+            let to_delete_vec = self
+                .deletion_scheduler
+                .write()
+                .pop_pending(pending_duration);
+            if !to_delete_vec.is_empty() {
```
javierhonduco: Two things here:

- Since you are checking if the vector isn't empty and then fetching the first element, why not match against it? The code would be cleaner that way
- Since `pop_pending` returns a vector, aren't we potentially not schuling pids `1..n`?

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
...
+                                .and_modify(|events| *events += 1)
+                                .or_insert(1);
+                        }
+                        // Now check how many mappings still exist for the PID, after
+                        // all should have been obliterated, and print out the keys in
+                        // debug format
+                        // - How many mappings per PID (mappings_by_pid HashMap)
+                        let mut mappings_by_pid: HashMap<u32, u32> = HashMap::new();
+                        // - Mappings we're going to have to manually delete
+                        let mut mappings_to_delete: Vec<_> = vec![];
```
javierhonduco: nit:
```suggestion
                        let mut mappings_to_delete = vec![];
```

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
...
+                        }
+                        // How bad were things?
+                        for (key, value) in mappings_by_pid {
+                            warn!(
+                                "Dead PID {} still has {} mappings! (will remove)",
+                                key, value
+                            );
+                        }
+                        // Now, delete those mappings
+                        for key in mappings_to_delete.iter() {
```
javierhonduco: Isn't this logic duplicated elsewhere? If this code needs to run perhaps it would be a symptom of a bigger issue that would be worth tackling.

src/bpf/tracers.bpf.c
```diff
@@ -49,9 +49,12 @@ int tracer_process_exit(void *ctx) {
     int per_process_id = BPF_CORE_READ(task, group_leader, thread_pid, numbers[level].nr);
     int per_thread_id = BPF_CORE_READ(task, thread_pid, numbers[level].nr);
 
-    if (!process_is_known(per_process_id)) {
-        return 0;
-    }
+    // To avoid a race condition with detecting and registering a PID *before*
+    // its exit tracepoint has fired, we just ignore whether we know about this
+    // PID or not, at least inside this BPF program
+    // if (!process_is_known(per_process_id)) {
```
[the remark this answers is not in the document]
gmarler: Fine by me

src/profiler.rs
gmarler: So, grab it once and use it forever?  If they ever make map's adjustable in size, I can imagine that could pose an issue.

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+                .info
+                .max_entries
+        );
+        // DEBUG exec_mappings usage:
+        // - Total PIDs represented (pids_with_mappings Vec)
+        let pids_with_mappings: Vec<_> = self
+            .native_unwinder
+            .maps
+            .exec_mappings
+            .keys()
```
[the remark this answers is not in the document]
gmarler: For now it's critical to track down what's going on with these mappings and should be on by default in all cases until we eliminate all of the corner cases.

It's also only done once a session, and over 5000 PIDs, so it should be _that bad_.

But I'm more than happy to wrap this in a timed log message so we can see if it's truly becoming a problem everywhere.

I do fully expect to remove this completely as we eliminate any final issues around mappings.

src/profiler.rs
gmarler: It doesn't have to be _that_ correct, as the DeletionScheduler makes PIDs "pending" for deletion after 2 sessions worth of time (10 seconds at the moment), so the resolution required is much more than a second.

src/profiler.rs
gmarler: I'll look at that, it's certainly possible I moved it from elsewhere and forgot to remove the source, which might account for one issue I'm seeing on rare occasions - certain mappings are already gone when we get around to deleting them.  Just note those and move on.

src/profiler.rs
```diff
@@ -2089,6 +2363,24 @@ impl Profiler {
     pub fn teardown_perf_events(&mut self) {
         self._links = vec![];
     }
+
+    fn live_pid_count(&mut self) -> usize {
+        let live_pids: Vec<Pid> = procfs::process::all_processes()
```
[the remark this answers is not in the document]
gmarler: Although we don't have any machines with more than a few 10s of thousands of processes (and less than 10 of those, if that) - others might though.

I'm not against gating this with an option, but we need this information now for all hosts as we go through beta, without thinking about turning it on - it just needs to be there.  Definitely agree on adding the option in 2026 though, as we get more stability overall.

Not seeing it be that expensive either at present - either in terms of CPU or memory.

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
...
+            debug!(
+                "First pass of pending_deletions had {} exited processes",
+                procs_to_reap
+            );
+            let procs = self.procs.read(); // read lock to start
+            pending_deletion.retain(|(pid, _)| procs.contains_key(pid));
+            // 2nd pass - Perform the actual final deletions
+            let pids_to_del_str = pending_deletion
+                .iter()
+                .map(|(n, _)| n.to_string())
```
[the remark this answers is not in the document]
gmarler: I'll give that a whirl

src/profiler.rs
```diff
@@ -852,41 +870,96 @@ impl Profiler {
...
+                .max_entries
+        );
+        // DEBUG exec_mappings usage:
+        // - Total PIDs represented (pids_with_mappings Vec)
+        let pids_with_mappings: Vec<_> = self
+            .native_unwinder
+            .maps
+            .exec_mappings
+            .keys()
+            .map(|key| exec_mappings_key::from_bytes(&key).unwrap().pid)
```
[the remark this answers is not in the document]
gmarler: I think I've got the way to `filter_map()` the Results of the `exec_mappings_key::from_bytes()` calls, while logging errors, and only keeping valid results.  See if you like it when I push the next commit.

src/bpf/tracers.bpf.c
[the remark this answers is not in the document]
javierhonduco: Let's think a bit more on the trade-offs of sending the event even if the process does not exist, as this will incur in more work for userspace

src/profiler.rs
javierhonduco: Yeah that's very confusing, given that `pop_pending` returns a single element, it should definitely not be a vector. I should have caught that during code review!

src/profiler.rs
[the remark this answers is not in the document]
javierhonduco: I am fine with adding this since it can be useful but definitely this should be opt-in. 

> Not seeing it be that expensive either at present - either in terms of CPU or memory.

With 10ks of processes this might be fast enough, but on a machine with 10x or 100x that number and under severe resource pressure this might become a hotspot. 

Either way, interested in seeing the flamegraph (at 997 Hz plz) for the machine you are testing this on. Would be good to see what's the performance cost of the current code for your workloads you are testing it on.

src/profiler.rs
```diff
@@ -2089,6 +2363,24 @@ impl Profiler {
     pub fn teardown_perf_events(&mut self) {
         self._links = vec![];
     }
+
+    fn live_pid_count(&mut self) -> usize {
+        let live_pids: Vec<Pid> = procfs::process::all_processes()
+            .expect("Cannot read proc")
```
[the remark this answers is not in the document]
javierhonduco: Sure, happy to help!

src/profiler.rs
```diff
@@ -2165,19 +2484,19 @@ mod tests {
...
+        assert!(
             profiler
                 .native_unwinder
                 .maps
                 .executable_to_page
                 .keys()
-                .count(),
-            0
+                .count()
+                > 0
```
javierhonduco: (I thought I had left a note here but perhaps I removed it by accident?)

This test needs an update to ensure that your changes are working properly and that this code doesn't break by accident. 

A possible approach would be to manually trigger a deletion after these asserts, and then assert that all the releveant data structures are, as expected, deleted.

src/profiler.rs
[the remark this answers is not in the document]
gmarler: Yah, I'll get that done today most likely.  But at this point performance takes a back seat to correctness.  The current git repo's code is demonstrably incorrect, as it runs out of memory in minutes/hours, and we're peeling the onion to get to a state where we can run lightswitch for weeks without worrying about that.

Being overly concerned about the performance at this stage seems to be premature optimization.

However, having the profile for future reference when making future changes makes good sense, especially as we improve _how_ this PR's code does what it does over time.

src/profiler.rs
gmarler: So, you want to store the _future_ Instant beyond which the PID deletion can occur, instead of the Instant at which the PID was added to the Deletion Scheduler?

Either works, of course, and I didn't see the API spelled out, so I went with what seemed intuitive.  Glad to pick one way or the other, as long as it doesn't prolong this PR merging.

src/profiler.rs
```diff
@@ -1788,7 +1868,11 @@ impl Profiler {
     }
 
     pub fn add_proc(&mut self, pid: Pid) -> Result<(), AddProcessError> {
+        // NOTE: There are 3 places where AddProcessError::ProcfsRace can be returned from this
```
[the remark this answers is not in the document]
gmarler: I tend to document ONLY in code, never in a PR - because I've never gone back and re-read a PR - not seeing the benefit as yet.

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
         Ok(())
     }
 
+    fn cleanup_procs(&mut self) {
+        // TODO: Wrap this function's activities in a logged duration timer
+        // Pop off any processes that we've kept around long enough after they've exited
+        // Where "long enough" is 2 sessions worth
+        let pending_duration = self.session_duration.checked_mul(2).unwrap();
```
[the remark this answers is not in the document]
gmarler: You don't have an option, from what I can see - `checked_mul()` is the only way to do a multiplication, and I want the code to be explicit about what and why we're doing this action.

And since both `session_duration` and 2 are constants (once the program is running), there's never a chance of a panic.

src/bpf/tracers.bpf.c
```diff
@@ -49,9 +49,12 @@ int tracer_process_exit(void *ctx) {
     int per_process_id = BPF_CORE_READ(task, group_leader, thread_pid, numbers[level].nr);
     int per_thread_id = BPF_CORE_READ(task, thread_pid, numbers[level].nr);
 
-    if (!process_is_known(per_process_id)) {
-        return 0;
-    }
+    // To avoid a race condition with detecting and registering a PID *before*
```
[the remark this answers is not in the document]
javierhonduco: Can you please clarify where you saw these logs? The path of the file or the command line you ran to see them would really help to understand what's really going on

gmarler:
Ok, here's the 60 second profile @997Hz of lightswitch built from this PRs current state, taken by another instance of lightswitch, but limiting to just to the originally running lightswitch (there's got to be a better way to explain that):

flame

This machine has a very small workload, but high short-lived process rates, as shown by the following final message over the 60 second period - had to look very closely to see any usage of `handle_process_exit()`  But obviously seeing lots of work in `cleanup_procs()`, but related to iterating over the `exec_mappings` map, as one might expect, which only happens for PIDs where we actually had mappings to begin with.

```
INFO report_resource_consumption: lightswitch::profiler: 4861 process exits this session; 32066 process exits since profiler startup
```

src/profiler.rs
```diff
@@ -2020,6 +2132,195 @@ impl Profiler {
...
     }
 
+    fn cleanup_procs(&mut self) {
+        // Wrap this function's activities in a logged duration timer
+        let _span = span!(Level::INFO, "cleanup_procs").entered();
+        // Pop off any processes that we've kept around long enough after they've exited
+        // Where "long enough" is 2 sessions worth
+        let pending_duration = self
+            .session_duration
+            .checked_mul(2)
```
javierhonduco: We agreed that if this can't overflow, there's no need for a checked multiplication, no?

src/profiler.rs
```diff
@@ -2020,6 +2132,195 @@ impl Profiler {
         Ok(())
     }
 
+    fn cleanup_procs(&mut self) {
+        // Wrap this function's activities in a logged duration timer
```
javierhonduco: nit: This comment is not totally correct, the function is not being wrapped. I would see that the comment is not needed as the behaviour of `span!` should be well understood for engineers perusing this codebase. We can always check the documentation if we have doubts

src/bpf/tracers.bpf.c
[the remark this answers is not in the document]
gmarler: Measured impact via all of:
* a `span!` (which I've since removed) in `handle_process_exit()` to see how high that latency can go
* The 997Hz profile that barely registers `handle_process_exit()` running at all - ran that on a few other machines other than the host with high short-lived process spikes
* And running on several workloads with very high process counts (tens of thousands), as well as very high short-lived process spikes, outside our test machine.

src/profiler.rs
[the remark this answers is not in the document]
gmarler: I can also add a `span!` for this, but also note that this isn't showing up at all in the 997Hz profile, so that's good.  Gating this with an option in a subsequent PR if it becomes a problem anywhere is very palatable as well.

src/profiler.rs
gmarler: Will definitely need either an example/detailed description of how to express such a test, or to discuss over Zoom via pairing.

src/profiler.rs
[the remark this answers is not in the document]
gmarler: confirmed - 997Hz profile doesn't show `report_resource_consumption()` at all with the most recent changes made in this PR.

src/profiler.rs
gmarler: I'm confident the way this works is correct, but need to discuss how to add a test so that if I make this an `if let...` I still have confidence that it's working.

src/profiler.rs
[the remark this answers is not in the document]
gmarler: Fixed this.

src/profiler.rs
gmarler: All fixed now.

src/profiler.rs
```diff
@@ -2020,6 +2112,188 @@ impl Profiler {
...
+                .deletion_scheduler
+                .write()
+                .pop_pending(pending_duration);
+            if !to_delete_vec.is_empty() {
+                let (pid, partial_write) = match to_delete_vec[0] {
+                    ToDelete::Process(_, pid, partial_write) => (pid, partial_write),
+                };
+                pending_deletion.push((pid, partial_write));
+            } else {
+                break;
```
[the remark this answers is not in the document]
gmarler: Yep, and this is working as expected now.

gmarler:
Made modifications to `cleanup_procs()` to only iterate through `exec_mappings` once per session, rather than once per dead PID mapping per session.

The 997 Hz profile looks quite a bit better:
bbperf-tt-774-flame-20251224

* `cleanup_procs()` looks to have been optimized away
* but I can see things it calls, like `pop_pending()` - just barely

gmarler:
And that's what it was.

@javierhonduco, I think you can re-review `cleanup_procs()` in particular, and this PR in general, because after the recent changes based on your first review:

* I barely see `cleanup_procs()` function at all in the 997 Hz profile (see the last profile above - still looks the same after all of the commits between then and this update).  Don't see `report_resource_consumption()` in the profile at all.
* Both procs and mappings are completely cleaned up with no leaks of mappings after many hours of operation after a couple of hiccups from my most recent changes - all fixed now
* I think I've cleaned up both all the issues you highlighted, and the nits - thanks much for those!
* Added span timing for both `cleanup_procs()` and `report_resource_consumption()` along the way, so we can explicitly know how much CPU they're actually consuming.  The former no longer gets anywhere close to using more than 100 ms per session (it was using up to 6 seconds (!!!) with the previous iteration pattern in the worst cases).  The latter no longer exceeds 18 ms per session.

gmarler (commented):
I think I've identified two places where mappings are being lost on process exit, and they're being reported by my last debug change.

After running for several hours, all lost mappings that need cleaning up are caused by one or the other.  So now we need to discuss how to prevent these corner cases, or at least clean up after them efficiently.

PR #370, opened 2026-01-22
wip wip wip

javierhonduco:
see: https://github.com/javierhonduco/lightswitch/pull/368

The draft PR above helped isolate the issue, then I verified that the logic we had regarding reference counting was correct, so use it rather than keep the state in two places.

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -807,6 +807,26 @@ impl Profiler {
                     debug!("collecting profiles on schedule");
                     let profile = self.collect_profile();
                     self.send_profile(profile);
+
+                    // @nocommit Minimal resource consumption report to verify cleanup is working
+                    // as expected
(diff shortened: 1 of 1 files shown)


javierhonduco:
I think CI is failing due to race conditions / errors spawning `/bin/bash`, will change this on Monday

src/profiler.rs
```diff
@@ -866,6 +866,15 @@ impl Profiler {
                 }
 
                 for mapping in &mut proc_info.mappings.0 {
+                    // Always delete mappings.
```
javierhonduco: We need to always delete the mappings from the BPF maps, not just when the reference count of the object files reaches 0.

We want to keep the reference counting logic but only for the unwind information data (pages & the unwind info itself)

PR #391, opened 2026-01-28
Show when we last got data during eviction

javierhonduco:
Show when we last got data during eviction for both processes and executables. This is intended to help understand the eviction decisions.

Test Plan
=========

CI


javierhonduco:
cc @samanpa

PR #393, opened 2026-01-29
chore: Fix typos in docs and comments

javierhonduco:
😬😬😬😬

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+Unreleased
+----------
+
+v0.3.1
+---------
+Initial documented release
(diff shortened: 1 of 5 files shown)


src/profiler.rs
```diff
@@ -1389,7 +1389,7 @@ impl Profiler {
             .iter()
         {
             // There is no unwind information for anonymous (JIT) mappings, so let's skip
-            // them. In the future we could either try to synthetise the unwind
+            // them. In the future we could either try to synthesise the unwind
```
gmarler: British/UK vs American/US usage.  Hadn't seen that particular one before

javierhonduco: Damn! The linter told me that synthetise was wrong and didn't ocur to me that it could be a UK variant, thought I just wrote it plain wrong... For a second got convinced I didn't know how to write synthetise. Will revert then 😝

PR #397, opened 2026-02-02
Bump version of lightswitch-proto

samanpa:
This version allows use of lightswitch-proto without installing `protoc`


javierhonduco:
Just pushed it https://crates.io/crates/lightswitch-proto/0.2.2

PR #446, opened 2026-03-24
Initial k8 support

xlanor:
Screenshot 2026-06-23 at 20 57 44

Screenshot 2026-03-24 at 23 56 33

Testing is done on a 6 node k3s cluster with some workloads I have that would be typical of any homelab. Note that the graphs reflect a total of all 6 lightswitch pods I have running that are deployed as DaemonSets

---

Initial cut at adding first class kube support for lightswitch.

This exposes a new `--kubernetes` mode that emits one pyroscope series per pod, in the form:

`lightswitch/node/namespace/owner/pod`.

From eBPF, we only get a pid, from kube, we only know the container ID. to map both to each other, use `/proc/<pid>/cgroup` to map eBPF samples back to the container ID from kube's api.

To avoid performing an expensive lookup each time against the api which returns all pod info, two metadata caches are maintained, container id -> pod_uid and pod_uid -> {namespace,name,owner} are kept with a max size of 3300 and 1100 respectively. Only if the lookup misses in both do we perform a call to kube api to find the new container ID.

A third cache (deferred_cids, cap 1024, 60s TTL) records misses so that on a cache miss we don't re-scan the api for the same unknown container until 60s later.

On the collector end, series are pushed by resolved pods. for RS, its collapsed to Deployment to avoid picking up stale series across of replicasets across rollouts.


docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM --platform=linux/amd64 rust:latest AS builder
```
javierhonduco: Let's add any additional thing such as dockerfiles separately

docker/README.md
```diff
@@ -0,0 +1,18 @@
+This dockerfile is for ease of use on ARM64 MacOS for compilation/running tests locally
+
+Specifically for the k8s feature, I develop and run tests on my mac mini, then ship to my k8s cluster for testing.
```
javierhonduco: I wonder if we should use [cargo-cross](https://github.com/cross-rs/cross) for this

lightswitch-metadata/src/containers/cid_resolver.rs
```diff
@@ -0,0 +1,230 @@
...
+
+pub struct CidResolver {
+    success_cache: Cache<i32, String>,
+    deferred_cache: Cache<i32, ()>,
+}
+
+impl CidResolver {
+    pub fn new() -> Self {
+        Self {
+            success_cache: Cache::builder()
```
javierhonduco, line 58: `lru` is not being used because of wanting to have a TTL, right?

lightswitch-metadata/src/containers/cid_resolver.rs
```diff
@@ -0,0 +1,230 @@
...
+                None
+            }
+        }
+    }
+}
+
+// Takes a pid and checks procfs to find a container id.
+fn pid_to_container_id(pid: i32) -> Option<String> {
+    let proc = Process::new(pid).ok()?;
+    let cgroups = proc.cgroups().ok()?;
```
javierhonduco, line 94: This is okay for a first iteration, but since accessing this might fail if the process exits before we read it we migth want to get the cgroup id from BPF directly and then resolve it somehow to what we need here

src/cli/main.rs
```diff
@@ -85,6 +86,13 @@ fn start_deadlock_detector() {
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
+    // kube-rs and reqwest both bring in rustls with different crypto providers
+    // (ring and aws-lc-rs). When both are present, rustls cannot auto-detect
+    // which to use. Explicitly install aws-lc-rs as the process-wide default
```
javierhonduco, line 105: ohh excellent, thank you!

src/cli/main.rs
```diff
@@ -284,15 +304,26 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                    &args.pyroscope_app_name,
+                    args.pyroscope_tenant_id.clone(),
+                    ProfilerConfig::default().session_duration,
+                    args.sample_freq,
+                    metadata_provider.clone(),
+                );
+                if args.kubernetes {
+                    Box::new(K8sCollector::new(
+                        metadata_provider.clone(),
+                        Box::new(pyro),
```
javierhonduco: This API design is a bit backwards, we should ideally use the collector and not send it around and add the k8s metadata extraction to the profiler itself as part of pluggable metadata providers

lightswitch-metadata/src/containers/k8s_metadata_cache.rs
```diff
@@ -0,0 +1,701 @@
...
+            .build();
+
+        let container_to_pod_clone = container_to_pod.clone();
+        let pod_metadata_clone = pod_metadata.clone();
+        let shutdown = Arc::new(AtomicBool::new(false));
+        let shutdown_clone = shutdown.clone();
+        let node_name_clone = node_name.clone();
+
+        // spawn background thread, run informer
+        let handle = std::thread::spawn(move || {
```
javierhonduco, line 71: Let's name the thread

lightswitch-metadata/src/containers/k8s_metadata_cache.rs
```diff
@@ -0,0 +1,701 @@
...
+                }
+            }
+        }
+    }
+}
+
+//  Parse the container ID from kube API
+//  kubectl get pod -n lightswitch lightswitch-h9dzp -o jsonpath='{.status.containerStatuses[0].containerID}'
+//  > containerd://36bbe89f0d3c9b982f920d6edf290b7083774993cca48b7fcf11e845875e8b4e%
+fn parse_container_id(raw: &str) -> Option<String> {
```
javierhonduco, line 357: Perhaps we can avoid the allocation here?

javierhonduco:
Pushed a version with my changes https://github.com/javierhonduco/lightswitch/tree/k8s-clean

xlanor:
@javierhonduco pulled off your commit, only issue there is the series. 

Nix is crapping out in ci, not sure if you could look at that

javierhonduco:
> Nix is crapping out in ci, not sure if you could look at that

No worries, it's a formatting issue, will fix in this PR

src/collector.rs
```diff
@@ -245,27 +248,75 @@ struct PyroscopeSample {
...
+
+    fn node_service_name(&self) -> String {
+        if self.node_name.is_empty() {
+            self.service_name.clone()
+        } else {
+            format!("{}/{}", self.service_name, self.node_name)
+        }
+    }
+
+    fn service_name_for_task(&self, pid: i32, tid: i32) -> String {
```
javierhonduco: This wasn't what I expected, what do you think about adding tags for namespace, pod, etc individually so we can group by them etc?

xlanor: Okay, let me know the final shape and I will implement it.

Mainly I think I mentioned before - the default pyro UI has no explorability for any of those, this will assume the user is always using pyro with grafana

javierhonduco: I think it should work fine, I did some testing when I implemented support for Pyroscope and sending the different labels should work. Perhaps one per k8s abstraction?

javierhonduco (approved):
LGTM thanks so much for your patience 🙇🏻 . I am not super familiar with k8s internals but this seems to make sense. Since this is an initial patch we can always work on improving it should any issues arise. 

As discussed off-GitHub, I've removed the Docker build bits since they are not being tested and they'll bitrot. I hear you here and want to provide an easier way for contributors to build lightswitch without having to have nix installed

javierhonduco:
The flaky tests that are failing are unrelated to this PR

Issue #394, opened 2026-01-30
Optimise `object::code_hash`


javierhonduco:
The current strategy of hashing the whole .text section for executables without a build, such as Rust's, id is quite CPU hungry as expected. Revisit this, either by using a different hashing function and / or not hashing the whole binary.

<img width="832" height="517" alt="Image" src="https://github.com/user-attachments/assets/43918d93-4bf4-4018-914a-835d92c8bfd6" />

javierhonduco:
A possible approach would be to hash:
- the first X KB of the executable
- the executable size
- the unwind info in `.eh_frame`, as this is smaller than `.text`?
- chunks of `.text`

Would be very useful to keep this as constant time as possible, since there are some really large binaries out there... Alternatively we could keep an on-disk cache for this, but feels a bit over engineered and it would have its own set of problems

PR #455, opened 2026-03-31
lightswitch-object: Remove .text hashing buffering

javierhonduco:
This is unnecessary and results in worse performance for no good reason. A generic reader is not needed since we have a byte slice.

Test Plan
=========

CI


javierhonduco:
Related to https://github.com/javierhonduco/lightswitch/issues/394

samanpa (approved):

PR #481, opened 2026-06-05
Read process name in BPF

javierhonduco:
Read process name in BPF
Before this commit, the process name and thread name were read in
userspace, in a best-effort basis. This is problematic as the process
name tends to be one of the most useful aggregation keys but it's not
always possible to read it off procfs due to the process exiting before
we get to read it.

Additionally, reading it from procfs is more expensive as in BPF is a
memory read (after a few pointer dereferences) and kernel locks are
grabbed etc.

As of now, the thread name is still read off procfs, so we might fail to
obtain it. In the future this might be revisited, and thread names might
also be collected from BPF.

Test Plan
=========

CI + manual tests (pprof and flamegraph generation)

Signed-off-by: Francisco Javier Honduvilla Coto <javierhonduco@gmail.com>

### DIFF
--- a/src/bpf/profiler.bpf.c
+++ b/src/bpf/profiler.bpf.c
@@ -68,7 +68,7 @@ struct {
 struct {
     __uint(type, BPF_MAP_TYPE_HASH);
     __uint(max_entries, MAX_PROCESSES);
-    __type(key, Event);
+    __type(key, u64);
     __type(value, bool);
 } rate_limits SEC(".maps");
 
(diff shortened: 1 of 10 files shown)


src/bpf/profiler.bpf.c
```diff
@@ -663,7 +663,9 @@ int on_event(struct bpf_perf_event_data *ctx) {
         .type = EVENT_NEW_PROCESS,
         .pid = per_process_id,
     };
-    bpf_get_current_comm(&event.comm, TASK_COMM_LEN);
+
+    // Send the main thread's name.
+    bpf_probe_read_kernel_str(&event.comm, TASK_COMM_LEN, &task->group_leader->comm);
```
samanpa: Should the retcode of this be checked?

javierhonduco: It shouldn't fail, this memory won't be swapped out, so it should always work

src/bpf/profiler.bpf.c
```diff
@@ -662,6 +663,9 @@ int on_event(struct bpf_perf_event_data *ctx) {
         .type = EVENT_NEW_PROCESS,
         .pid = per_process_id,
     };
+
+    // Send the main thread's name.
+    BPF_CORE_READ_STR_INTO(&event.comm, task, group_leader, comm);
```
samanpa, line 667: Should the error code be checked?

samanpa (approved):

PR #496, opened 2026-06-22
Add Ability to Activate per Sample Timestamps

gmarler:
Currently, lightswitch applies one timestamp per session (every 5 seconds, by default) across all samples collected during the session window.  In the case of creating a pprof per session and shipping the result to lightswitch-backend, the timestamp is applied upon actual ingestion time, not collection time.

This PR has the following rationale:
* This feature is behind a flag, which is false by default: `--enable-ts-per-sample`
* When enabled, apply a timestamp to every sample collected, much closer to the moment of collection
* As a practical matter, nanosec/microsec resolution timestamps are too fine grained, so the timestamps collected are millisec resolution.  From past experience, even sub-second profiling generally buckets samples on a per 1-100 ms resolution, so millisec resolution should be perfectly adequate
* The timestamp is encoded in a new pprof label: `collected_at`. This will allow the backend to continue operating as-is until support is added for the new label at that level
* This feature allows much finer grained (down to sub-second, eventually) zooming in of profiling time ranges, but even at per-second resolution is quite useful to zero in on shorter lived code path activity
* Tests have been added 

A separate POC has already been performed, adding support for these timestamps in lightswitch-backend and the UI, just to confirm this approach is functional, and achieves the stated goals.


src/bpf/profiler.bpf.c
```diff
@@ -264,7 +264,7 @@ static __always_inline void add_stack(struct bpf_perf_event_data *ctx,
 
     unwind_state->sample.pid = per_process_id;
     unwind_state->sample.tid = per_thread_id;
-    unwind_state->sample.collected_at = bpf_ktime_get_boot_ns();
+    unwind_state->sample.collected_at = bpf_ktime_get_ns();
```
javierhonduco: Why was this helper changed? We would like to account for the time the computer has been sleeping, no?

src/cli/args.rs
```diff
@@ -185,6 +185,11 @@ pub(crate) struct CliArgs {
         help = "Read metadata for all threads when a new process is detected. This might be slow in older kernels."
     )]
     pub(crate) preload_thread_metadata: bool,
+    #[arg(
+        long,
+        help = "Attach a per-sample collected_at timestamp (ms) label to pprof output"
+    )]
+    pub(crate) enable_ts_per_sample: bool,
```
javierhonduco, line 192: We should maybe emit a warning if the profile format is not pprof and this is enabled since other formats support high resultion timestamps already (see the firefox profiler collector)

src/profile/sample.rs
```diff
@@ -329,6 +332,74 @@ mod tests {
         );
     }
 
+    fn hash_of(sample: &RawSample) -> u64 {
```
javierhonduco: Should this impement Hash/er directly?

javierhonduco:
Tangentially related but I was thinking of adding support for [OTel](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto) which supports "aggregated" profiles à la pprof and also timestamped samples

src/bpf/profiler.bpf.c
gmarler: Because that helper wasn't available for at RHEL 8.10's kernel, whereas the one I'm using is available universally.

I was surprised at this.

And yes, sleep time isn't accounted for, but the intervals are so small that it's not (currently) an issue.

Would you prefer I add a config flag (like `use_boot_ns`, for instance) similar to other flags you have in `lightswitch_config` already so that kernels that have it could use `bpf_ktime_get_boot_ns()`, and older kernels like RHEL 8.10 can fall back to `bpf_ktime_get_ns()`?

src/cli/args.rs
gmarler: Fine - do you not expect to support those timestamp types as well, in possible future PRs?

src/profile/sample.rs
gmarler: Suppose so.  I'll see how involved (or easy) it is.

gmarler:
> Tangentially related but I was thinking of adding support for [OTel](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/profiles/v1development/profiles.proto) which supports "aggregated" profiles à la pprof and also timestamped samples

Certainly doable, but one format per PR, eh?

src/profile/sample.rs
gmarler: So, would you prefer this as a trait implementation (which would only be for the tests), or to alternatively use a simpler `BuildHasher` variant?

I'm doing the trait first, but let me know if you want me to switch.

src/bpf/profiler.bpf.c
gmarler: It wasn't that hard to auto-detect whether the capability exists, just as you've done for other things, so I did that.

src/cli/args.rs
gmarler: So the warning would be (currently only for the firefox profiler collector format, and any others you may add in future): "--enable-ts-per-sample is not necessary for output format X - such timestamps are provided by default"?

Or what would you prefer?

src/bpf/profiler.bpf.c
gmarler: BUT... after making that change, something interesting happened - 2 tests started to get verifier failures, but the binary worked just fine.

I explain what I had to do to get rid of that in the commit message for 6d0be10.

What's different now is that the flags default to false, but the check against system info will set them to the correct detected values - apparently the binary already did this, but the tests did not.  Fine for the one flag that's apparently available on every kernel we've ever used it with, but the flag I just added isn't always available.

Setting _any_ of these flags to true by default gives the verifier the chance to check for the non-present features, which the verifier apparently doesn't like (and probably only on the older kernels like RHEL 8.10).

Convoluted!

javierhonduco: Beyond the fact that lightswitch itself runs without issues on this distro version, additional proof that RedHat 8.10 has support for `bpf_ktime_get_boot_ns()`:

```
# cat /etc/os-release | grep 8.10
VERSION="8.10 (Ootpa)"
VERSION_ID="8.10"
PRETTY_NAME="Red Hat Enterprise Linux 8.10 (Ootpa)"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.10
REDHAT_SUPPORT_PRODUCT_VERSION="8.10"
```


```
#  sudo cat /proc/kallsyms |& grep bpf_ktime_get_boot_ns
ffffffff9b461c60 T bpf_ktime_get_boot_ns
ffffffff9c237b80 R bpf_ktime_get_boot_ns_proto
```

As you mentioned, the tests hardcode certain settings that might not be available, that's the ultimate root cause of the issue while running the tests on older kernels, but this all seems unrelated to me to the `bpf_ktime_get_boot` helper. I think we should not have this feature detection since all the kernels we support [should have it](https://docs.ebpf.io/linux/helper-function/bpf_ktime_get_boot_ns/) (RHEL 8 [backported it, it seems](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html-single/8.10_release_notes/index)).

src/cli/args.rs
javierhonduco: Let me think about this, need to make sure we don't under/over design this feature

lightswitch-proto/src/profile.rs
```diff
@@ -478,6 +478,43 @@ mod tests {
...
+        assert_eq!(sample.label.len(), 1);
+        let lbl = &sample.label[0];
+        assert_eq!(lbl.num, ms_timestamp);
+        assert_ne!(lbl.num_unit, 0, "num_unit must be set for numeric labels");
+        let key_str = &profile.string_table[lbl.key as usize];
+        assert_eq!(key_str, "collected_at");
+        let unit_str = &profile.string_table[lbl.num_unit as usize];
+        assert_eq!(unit_str, "milliseconds");
+
+        // Verify after encode/decode round-trip
```
javierhonduco, line 508: This test is making sure that the (de)serialization works, we don't need to check for this, we can assume it works and that upstream is well tested

src/bpf/profiler.bpf.c
```diff
@@ -264,7 +264,9 @@ static __always_inline void add_stack(struct bpf_perf_event_data *ctx,
 
     unwind_state->sample.pid = per_process_id;
     unwind_state->sample.tid = per_thread_id;
-    unwind_state->sample.collected_at = bpf_ktime_get_boot_ns();
+    unwind_state->sample.collected_at = lightswitch_config.use_ktime_get_boot_ns
+        ? bpf_ktime_get_boot_ns()
+        : bpf_ktime_get_ns();
```
javierhonduco: Let's back out this change. [This helper is always present ](https://github.com/javierhonduco/lightswitch/pull/496#discussion_r3503971512)

src/bpf/profiler.bpf.c
gmarler: Yes, `bpf_ktime_get_boot_ns` has been there all the time, it was `task_pt_regs_helper` that was missing all along on RHEL 8.10, and I mentally switched them.

So `bpf_ktime_get_boot_ns` doesn't _need_ a detector at present (personally, I'd have it anyway, just so everything is consistent), so I'll remove it.

The genesis of this issue seems to have been that these tests have been failing for quite some time on RHEL 8.10, but since we don't explicitly run the test suite as part of generating the package, this was simply never noticed.  

The binary auto-detects whether the helper is available, so the problem never shows up there - I just assumed that if the tests died with a verifier error, the binary surely would, and never tried it.

lightswitch-proto/src/profile.rs
gmarler: Did you mean:
* // Verify in memory
* // Verify after encode/decode round-trip

Or both?  Or basically the entire test?

javierhonduco:
As discussed elsewhere, rather than retrofitting per sample timestamps in pprof, this would be a great candidate for adding OTel profiling as it natively supports this use case

PR #499, opened 2026-06-25
Prepare release

javierhonduco:


lightswitch-metadata/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "lightswitch-metadata"
-version = "0.2.2"
+version = "0.3.0"
```
samanpa: Are there any API changes? If not maybe stay on 0.2.3. Less of a hassle to upgrade.

samanpa (approved):

lightswitch-metadata/Cargo.toml
javierhonduco: Not really, we can stay on 0.2.3 without issues

PR #503, opened 2026-07-01
Unify env variable to enable libbpf logs in tests

gmarler:
- This allows getting BPF verifier debug output when such tests fail

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -153,7 +153,7 @@ impl Default for ProfilerConfig {
     fn default() -> Self {
         Self {
             cache_dir_base: temp_dir(),
-            libbpf_debug: std::env::var("LIBBPF_DEBUG").is_ok(),
+            libbpf_debug: false,
             bpf_logging: false,
             duration: Duration::MAX,
             sample_freq: 19,
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -153,7 +153,7 @@ impl Default for ProfilerConfig {
     fn default() -> Self {
         Self {
             cache_dir_base: temp_dir(),
-            libbpf_debug: false,
+            libbpf_debug: std::env::var("LIBBPF_DEBUG").is_ok(),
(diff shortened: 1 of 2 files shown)


src/profiler.rs
```diff
@@ -153,7 +153,7 @@ impl Default for ProfilerConfig {
     fn default() -> Self {
         Self {
             cache_dir_base: temp_dir(),
-            libbpf_debug: false,
+            libbpf_debug: std::env::var("LIBBPF_DEBUG").is_ok(),
```
javierhonduco: This is false by default as it's a reasonable setting to have unless something it's set on purpose. 

I recommend to either manually add it for each test such as it's done in https://github.com/javierhonduco/lightswitch/blob/60048b38fb66cc090e6b31db258c7e05f939e0dc/tests/integration_test.rs#L323, or that a a constructor for testing purposes is created

gmarler: Sure, can do that - but this keeps the value false by default (and has the benefit of only having to be set in one place with one change - which I thought was a virtue) - one has to set `LIBBPF_DEBUG` to get a value other than false.

Relatedly, did you want me to change `LIBBPF_DEBUG` to `TEST_DEBUG_BPF`, for consistency across the codebase?  Or was the intent different?

javierhonduco: I think it's too "magical" to use env vars for things like this, that's why I prefer hardcoding it / selecting it at runtime. The reason for this is that typically env vars are not very well documented.

I think `TEST_LIBBPF_DEBUG` makes more sense, definitely would be good to have a single env var name for this purpose. But as long as we are consistent on using one name I don't mind too much which one we choose!

gmarler: Will do!

gmarler: Ok, I think I've finished here - please confirm/deny.

javierhonduco (approved):
LGTM, thank you!

PR #504, opened 2026-07-01
Fix BPF tests failing because use_task_pt_regs_helper defaults to true

gmarler:
* Fix the tests that fail because use_task_pt_regs_helper defaults to `true`
* Once those pass, subsequent integration tests fail on RHEL 8.10 - fix those

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -1556,7 +1556,18 @@ mod tests {
             &profiler.bpf.native_unwinder.maps
         }
 
-        let mut profiler = Profiler::default();
+        let (_stop_signal_send, stop_signal_receive) = crossbeam_channel::bounded(1);
+        let metadata_provider = std::sync::Arc::new(std::sync::Mutex::new(
+            lightswitch_metadata::metadata_provider::GlobalMetadataProvider::default(),
+        ));
+        let mut profiler = Profiler::new(
+            ProfilerConfig {
+                use_task_pt_regs_helper: false,
+                ..ProfilerConfig::default()
+            },
+            stop_signal_receive,
+            metadata_provider,
+        );
 
         // All BPF maps must be empty.
         assert_eq!(maps(&profiler).exec_mappings.keys().count(), 0);
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -167,6 +167,8 @@ fn test_integration() {
         duration: Duration::from_secs(5),
         sample_freq: 999,
         userspace_pid_ns_level: system_info.available_bpf_features.userspace_pid_ns_level,
+        use_task_pt_regs_helper: system_info.available_bpf_features.has_task_pt_regs_helper
+            && system_info.available_bpf_features.has_get_current_task_btf,
         ..Default::default()
     };
     let (_stop_signal_send, stop_signal_receive) = bounded(1);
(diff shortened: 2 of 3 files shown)


gmarler:
Now that those tests pass, the following fail:

```
failures:

---- test_use_pt_regs_helper stdout ----

thread 'test_use_pt_regs_helper' (190377) panicked at src/bpf_objects.rs:70:66:
load skel: Error: Invalid argument (os error 22)

---- test_do_not_use_pt_regs_helper stdout ----

thread 'test_do_not_use_pt_regs_helper' (190374) panicked at src/bpf_objects.rs:70:66:
load skel: Error: Invalid argument (os error 22)

---- test_integration_ocaml_native_defaults stdout ----

thread 'test_integration_ocaml_native_defaults' (190376) panicked at src/bpf_objects.rs:70:66:
load skel: Error: Invalid argument (os error 22)

---- test_integration stdout ----

thread 'test_integration' (190375) panicked at src/bpf_objects.rs:70:66:
load skel: Error: Invalid argument (os error 22)


failures:
    test_do_not_use_pt_regs_helper
    test_integration
    test_integration_ocaml_native_defaults
    test_use_pt_regs_helper
```

gmarler:
With that commit, the integration tests pass again on a live RHEL 8.10 box too.

src/profiler.rs
```diff
@@ -1556,7 +1556,18 @@ mod tests {
...
+        let metadata_provider = std::sync::Arc::new(std::sync::Mutex::new(
+            lightswitch_metadata::metadata_provider::GlobalMetadataProvider::default(),
+        ));
+        let mut profiler = Profiler::new(
+            ProfilerConfig {
+                use_task_pt_regs_helper: false,
+                ..ProfilerConfig::default()
+            },
+            stop_signal_receive,
+            metadata_provider,
```
javierhonduco, line 1569: Why not using `..Profiler::Default()` like you did above for the config?

tests/integration_test.rs
```diff
@@ -330,7 +340,7 @@ fn test_do_not_use_pt_regs_helper() {
         libbpf_debug: bpf_test_debug,
         bpf_logging: bpf_test_debug,
         duration: Duration::from_millis(100),
-        use_task_pt_regs_helper: true,
+        use_task_pt_regs_helper: false,
```
javierhonduco, line 344: Seems like this was the original intent, thanks for changing it!

tests/integration_test.rs
```diff
@@ -299,6 +303,12 @@ fn test_integration_ocaml_native_defaults() {
 #[test]
 fn test_use_pt_regs_helper() {
     let bpf_test_debug = std::env::var("TEST_DEBUG_BPF").is_ok();
+    let system_info = SystemInfo::new(None).expect("failed to detect system info");
+    if !system_info.available_bpf_features.has_task_pt_regs_helper
+        || !system_info.available_bpf_features.has_get_current_task_btf
+    {
+        return;
```
javierhonduco, line 311: Let's print that this feature is not available and the tests are being skipped

javierhonduco (approved):
Minor comments, LGTM overall

src/profiler.rs
gmarler: Oh, that's because here we're not overriding `ProfilerConfig::default()` (alone) like I did above, but I'm overriding `Profiler::default()` instead - because that's what the test is using.

And the only way I know of to do that in Rust is not to call `Profiler::Default()` at all, but to call `Profiler::new()` with the same `ProfilerConfig::default()` override as I did above - plus the other arguments `Profiler::new()` needs.

Am I missing something?

tests/integration_test.rs
gmarler: Sure thing.

tests/integration_test.rs
gmarler: Done.

gmarler:
@javierhonduco , FYI, I don't have the option to merge these changes apparently, even though you approved them.  Could you do the honors?

gmarler:
Nevermind - the UI simply didn't update to say you had already merged the PR!

PR #495, opened 2026-06-22
Add unroll to dwarf_unwind outer loop

gmarler:
This seems to be necessary for a verifier regression introduced in ~6.8+ kernels, which we see in later RHEL 8.10 kernel updates.

Without it, the verifier rejects the eBPF program in those later releases.

PR #506, opened 2026-07-06
build: Use clang/llvm 21

javierhonduco:
Use clang 21, initially released on the 26 August 2025 (1y after clang 19).

The generated code trips the verifier in older kernels we support, after bisecting clang, the offending commit is `[BPF] Make -mcpu=v3 as the default (#107008) ` [0]. The code generated with this version of the ISA isn't too different, but some registers seem to be 16 and 32 bit wide instead of 64. Perhaps the verifier can't prove termination anymore due to this. Forcing these values to be 64 bits seems like a bit of a hack, so let's force v1 of the ISA and the ISA bump can be done later on.

Note that unrolling the loop as previously suggested [1] makes the code size explode, as expected:

```
$ wc -l /tmp/llvm_21_v3 /tmp/llvm_21_v1 /tmp/llvm_21_v1_unrolled
  1748 /tmp/llvm_21_v3
  1800 /tmp/llvm_21_v1
  7948 /tmp/llvm_21_v1_unrolled
 11496 total
```

[0]: https://github.com/llvm/llvm-project/commit/7852ebc088b925ef1c1940cbd56a93d9f8e3e330
[1]: https://github.com/javierhonduco/lightswitch/pull/495

Test Plan
=========

CI

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -65,7 +65,9 @@ fn main() {
     let skel = Path::new(PROFILER_SKELETON);
     SkeletonBuilder::new()
         .source(PROFILER_BPF_SOURCE)
-        .clang_args(["-Wextra", "-Wall", "-Werror"])
+        // Older kernels reject the current code if compiled with v3 of the ISA (the default in
+        // LLVM 20+). Select the previous default BPF instruction set.
+        .clang_args(["-mcpu=v1", "-Wextra", "-Wall", "-Werror"])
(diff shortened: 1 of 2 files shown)


javierhonduco:
cc @gmarler @samanpa

samanpa (approved):
This is great thanks

PR #520, opened 2026-07-26
cli: Expose Firefox Profiler format properly

javierhonduco:
The current code is a bit silly, but let's at least expose the firefox profiler format as what it is, a format, rather than the leaky messy code of being a separate collector.

Test Plan
=========

CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,7 @@ As a CLI, **lightswitch** can be run with:
 $ sudo lightswitch
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -34,13 +34,7 @@ As a CLI, **lightswitch** can be run with:
 $ sudo lightswitch
 ```
 
-Stop it with <kbd>Ctrl</kbd>+<kbd>C</kbd>, or alternatively, pass a `--duration` in seconds. By default, a flamegraph in SVG format will be written to disk. Pprof is also supported with `--profile-format=pprof`. By default the whole machine will be profiled which can be overriden with `--pids`.
-
-With Docker:
-
-```shell
-$ docker run -it --privileged --pid=host -v /sys:/sys -v $PWD:/profiles -v /tmp/lightswitch ghcr.io/javierhonduco/lightswitch:main-$LIGHTSWITCH_SHA1 --profile-path=/profiles
```
javierhonduco, line 42: This doesn't really belong here, will re-add it in the future

PR #526, opened 2026-07-27
profile: Add Perfetto trace output and viewer integration

### LINKED ISSUE #521: Support for Perfetto output
Hi Javier!

I came across Lightswitch while doing some research for [sismo](https://github.com/LalitMaganti/sismo). I was wondering if you were interested in adding Perfetto protobuf output to lightswitch as wellt to visualize profiles?

I recently defined a format upstream (https://github.com/google/perfetto/discussions/6027) which allows to represent timestamped stack sampling information using protobuf. Spiritually I think it would be very similar to how the [Firefox collector](https://github.com/javierhonduco/lightswitch/blob/62fc32b4ce5a36d2693973a65b65649007743e3f/src/collector.rs#L539) works.

Happy to send a PR if interested :)

LalitMaganti:
Add Perfetto as a first-class CLI profile format for investigations where aggregated stacks lose temporal context. A dedicated collector preserves each sample's CLOCK_BOOTTIME timestamp, optionally symbolizes it locally, and writes interned StackSample packets without generated Perfetto protobuf bindings.

Generalize `lightswitch server` to open completed Firefox or Perfetto profiles in their native web viewers. It selects the sole default output automatically, rejects missing or ambiguous defaults, and still accepts explicit paths and format overrides. Tokenized CORS URLs and tower-http's ranged file serving make large traces practical while removing the direct tokio-util dependency.

Fixes: https://github.com/javierhonduco/lightswitch/issues/521

### DIFF
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -1,5 +1,6 @@
 use core::str;
 use std::error::Error;
+use std::ffi::CString;
 use std::fs::File;
 use std::io::IsTerminal;
 use std::io::Write;
@@ -15,14 +16,13 @@ use clap::Parser;
 use crossbeam_channel::bounded;
 use crossbeam_channel::tick;
 use inferno::flamegraph;
-use lightswitch::collector::FirefoxProfilerCollector;
 use lightswitch::collector::{
-    AggregatorCollector, Collector, LiveCollector, NullCollector, PyroscopeCollector,
-    StreamingCollector,
+    AggregatorCollector, Collector, FirefoxProfilerCollector, LiveCollector, NullCollector,
+    PerfettoCollector, PyroscopeCollector, StreamingCollector,
 };
 use lightswitch::debug_info::DebugInfoManager;
 use lightswitch::profile::symbolize_profile;
-use nix::unistd::Uid;
+use nix::unistd::{Gid, Uid};
 use tracing::{debug, error, info, Level};
 use tracing_subscriber::fmt::format::FmtSpan;
 use tracing_subscriber::FmtSubscriber;
@@ -39,7 +39,7 @@ use lightswitch::debug_info::{
 use lightswitch::kernel::kernel_build_id;
 use lightswitch::profile::{fold_profile, to_pprof};
 use lightswitch::profiler::{Profiler, ProfilerConfig};
-use lightswitch::server::start_server;
+use lightswitch::server::{find_default_profile, start_profile_server, ProfileFileFormat};
 use lightswitch_object::kernel::kaslr_offset;
 use lightswitch_object::ObjectFile;
 use lightswitch_unwind_info::compact_unwind_info;
@@ -56,6 +56,7 @@ use crate::args::FlamegraphAggregation;
 use crate::args::LoggingLevel;
 use crate::args::ProfileFormat;
 use crate::args::ProfileSender;
+use crate::args::ServerProfileFormat;
 use crate::args::Symbolizer;
 use crate::killswitch::KillSwitch;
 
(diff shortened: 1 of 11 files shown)


javierhonduco (commented):
Thank you! Will review later today. Just one thing, noticed that this is serialising data to Protocol buffers by hand rather than using a code generator, [like we do for pprof](https://github.com/javierhonduco/lightswitch/tree/main/lightswitch-proto). I don't mind either way, since the code seems clean enough but I am wondering if this could hinder maintainability, as it has to be kept up to date if the proto upstream changes.

Let me know what you think

LalitMaganti:
> Just one thing, noticed that this is serialising data to Protocol buffers by hand rather than using a code generator, [like we do for pprof](https://github.com/javierhonduco/lightswitch/tree/main/lightswitch-proto)

Yes this was an intentional choice on my part: in general, over the years we've found general-purpose protobuf libraries (especially the official Google ones) sub-optimal for encoding large streams of protos like Perfetto traces. Specifically, because most of them represent protos using an "object model", you end up with double copying bytes and also allocating everywhere unnecessarily. Which is why we've always ended up implementing direct bytestream encoders instead. 

Now there is a middle ground where you can get a code generator to generate the field ids and bytestream encoding wrappers instead; now I don't know if there's a Rust library out there that does that (or if prost can me made to work in that way) but happy to switch to that if you prefer.

> but I am wondering if this could hinder maintainability, as it has to be kept up to date if the proto upstream changes.

I would very much *not* worry about this :) The field ids we're using here are never going to change and we're pretty strict about maintaining backcompat in Perfetto (most traces even from 2017 should open *just fine* in Perfetto UI today).

javierhonduco:
Makes sense! I guessed there was a reason behind this but wanted to make sure :)

javierhonduco:
While testing I noticed that the profile I captured looked a bit odd

image

And there was this error in the UI


```
UI: https://ui.perfetto.dev/v57.2-da1d152cf

RangeError: The number NaN cannot be converted to a BigInt because it is not an integer
 - BigInt (<anonymous>)
 - new _HighPrecisionTime (../../src/base/high_precision_time.ts:46:4)
 - _HighPrecisionTime.addNumber (../../src/base/high_precision_time.ts:116:4)
 - TimeScale.pxToHpTime (../../src/base/time_scale.ts:52:4)
 - Object.onDrag (../../src/core_plugins/dev.perfetto.Timeline/minimap.ts:118:10)
 - Object.onDrag (../../src/core_plugins/dev.perfetto.Timeline/minimap.ts:341:16)
 - HTMLDivElement.onPointerMove (../../src/base/mithril_utils.ts:422:6)

Trace: not available (FILE). Provide repro steps.
UA: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36
Referrer: http://127.0.0.1:9001/
```

The captured pftrace (compressed as zip so GitHub would let me upload it)
[profile.pftrace.zip](https://github.com/user-attachments/files/30499679/profile.pftrace.zip)

LalitMaganti:
> While testing I noticed that the profile I captured looked a bit odd

Ahhh I didn't clarify: you need to open this in this in the autopush UI, this is *very* newly landed upstream. Go to flags on sidebar -> release chanell -> autopush

On my machine:

image

> And there was this error in the UI

Ahhh I think this is this is a zero duration trace: we should handle this more gracefully! Filed https://github.com/google/perfetto/issues/6915 upstream for this.

javierhonduco:
> Ahhh I didn't clarify: you need to open this in this in the autopush UI, this is very newly landed upstream. Go to flags on sidebar -> release chanell -> autopush


Ohh sweet, will try again soon!

src/cli/main.rs
```diff
@@ -157,25 +158,50 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-                    .expect("SUDO_UID not set")
-                    .parse::<u32>()
-                    .expect("SUDO_UID could not be parsed");
-
-                nix::unistd::setuid(Uid::from_raw(target_uid)).unwrap();
+                let target_uid = Uid::from_raw(std::env::var("SUDO_UID")?.parse()?);
+                let target_gid = Gid::from_raw(std::env::var("SUDO_GID")?.parse()?);
+                let target_user = CString::new(std::env::var("SUDO_USER")?)?;
+                nix::unistd::initgroups(&target_user, target_gid)?;
+                nix::unistd::setgid(target_gid)?;
```
javierhonduco, line 173: Ohh nice one

src/cli/main.rs
```diff
@@ -459,7 +505,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 }
             }
         }
-        ProfileFormat::Firefox => {} // Handled separately, as a different collector
+        ProfileFormat::Firefox | ProfileFormat::Perfetto => unreachable!("handled above"),
```
javierhonduco, line 508: Way better! Thank you for making these improvements

javierhonduco (approved):
LGTM!! Thank you!! Just pushed a lint fix, rebased, and added this file format to the gitignore

PR #527, opened 2026-07-27
unwind: Handle large stack frames

### LINKED ISSUE #522: Stacks break for functions with large stack frames
The unwind tables store the size of each function's stack frame in a
16-bit field. The code that writes the tables treats that field as
unsigned, but the BPF code that reads them treats it as signed. Frame
sizes between 32 KiB and 64 KiB come back negative, so the unwinder
walks in the wrong direction and the stack is lost. A function only
needs a ~40 KiB local buffer to hit this.

## Reproduce

```c
// bigframe.c
#include <stdio.h>
__attribute__((noinline)) unsigned long hot(unsigned long x) {
  volatile unsigned char buf[40000];
  for (int i = 0; i < 40000; i++) buf[i] = (unsigned char)(x + i);
  for (int i = 0; i < 40000; i++) x += buf[i];
  return x;
}
int main(void) {
  unsigned long a = 0;
  for (;;) a += hot(a);
}
```

```sh
clang -O1 -fomit-frame-pointer -fasynchronous-unwind-tables -o bigframe bigframe.c
lightswitch show-unwind ./bigframe | grep 39888
```

The row for `hot()` shows `cfa_offset: 39888`. Read as signed that is
-25648, so profiling this program produces no usable stacks for `hot()`.

----

Happy to send a PR for this! Fix is quite simple :)

LalitMaganti:
Fix the signed -> unsigned silent conversion and add tests to catch these sort of issues in the future.

Fixes: https://github.com/javierhonduco/lightswitch/issues/522

### DIFF
--- a/src/bpf/profiler.bpf.c
+++ b/src/bpf/profiler.bpf.c
@@ -398,7 +398,7 @@ int dwarf_unwind(struct bpf_perf_event_data *ctx) {
         u64 found_pc = object_relative_pc_high + row->pc_low;
         u8 found_cfa_type = row->cfa_type;
         u8 found_rbp_type = row->rbp_type;
-        s16 found_cfa_offset = row->cfa_offset;
+        u16 found_cfa_offset = row->cfa_offset;
         s16 found_rbp_offset = row->rbp_offset;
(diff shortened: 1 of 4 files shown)


javierhonduco (approved):
Great find! Thank you very much!!

PR #528, opened 2026-07-27
unwind: Support sectionless ELF binaries

### LINKED ISSUE #525: Binaries with stripped section headers can't be profiled
Section headers are optional in ELF binaries — some build pipelines
strip them and the programs run fine. lightswitch finds the unwind data
by section name, so on such a binary it finds nothing, even though the
data is still in the file and the program headers say exactly where it
is. The `show-unwind` command panics outright.

## Reproduce

```sh
echo 'int main(void){for(;;){}}' > hello.c
clang -O2 -fasynchronous-unwind-tables -o hello hello.c
llvm-objcopy --strip-sections hello sectionless

lightswitch show-unwind ./sectionless
```

```
thread 'main' panicked at src/cli/main.rs:504:55:
called `Result::unwrap()` on an `Err` value: no .eh_frame section found
```

Profiling such a binary fails the same way, and the process ends up on
the afflicted list with no stacks. `readelf -l sectionless` shows a
`GNU_EH_FRAME` program header pointing at the unwind data the whole
time.

---

Happy to send a PR for this as well. The fix for this is to locate the unwind
data through the and have the CLI report the error instead of panicking.

LalitMaganti:
## Summary

- fall back to `PT_GNU_EH_FRAME` when an ELF file has no `.eh_frame` section header
- decode `.eh_frame_hdr` and map its `.eh_frame` address through the containing load segment
- isolate and document the ELF-specific resolution logic
- propagate `show-unwind` errors instead of panicking
- add a sectionless ELF regression fixture

Fixes #525.

## Test Plan

- `cargo test -p lightswitch-unwind-info`
- `cargo test --no-default-features --bin lightswitch`
- `cargo clippy --workspace --all-targets --no-default-features -- -D warnings`
- manually verified `lightswitch show-unwind` emits unwind rows for the sectionless fixture


lightswitch-unwind-info/src/elf.rs
```diff
@@ -0,0 +1,120 @@
...
+) -> Result<ResolvedUnwindInfo<'data>>
+where
+    Header: ProgramHeader<Endian = Endianness>,
+{
+    // PT_GNU_EH_FRAME contains `.eh_frame_hdr`, not `.eh_frame`. Decode it to
+    // find the virtual address of the actual unwind data.
+    let header = headers
+        .iter()
+        .find(|header| header.p_type(endian) == PT_GNU_EH_FRAME)
+        .ok_or(UnwindInfoError::NoEhFrameSection)?;
```
javierhonduco: nit: Perhaps being more specific here could help debugging this code if there's any issues
```suggestion
        .ok_or(UnwindInfoError::NoEhFrameHeader)?;
```

javierhonduco:
This is pretty awesome, I'll test it on some _odd_ binaries I have lying around just to ensure this works well on other compilers, runtimes, and compiler options

lightswitch-unwind-info/src/elf.rs
LalitMaganti: Done thanks!

javierhonduco (approved):
Thank you!

Issue #524, opened 2026-07-27
Go: small hot functions show up under the wrong caller


LalitMaganti:
For Go binaries, lightswitch assumes every function keeps a frame
pointer and unwinds through it. But Go gives no frame at all to small
leaf functions. When a sample lands in one, following the frame pointer
skips the function's real caller, and the profile shows the hot
function attached to its grandparent instead.

## Reproduce

```go
// leafgo.go
package main

//go:noinline
func leaf(x uint64) uint64 {
        for i := uint64(0); i < 4096; i++ {
                x = x*2654435761 + i
        }
        return x
}

//go:noinline
func mid(x uint64) uint64 {
        for i := 0; i < 8; i++ {
                x = leaf(x)
        }
        return x
}

func main() {
        var a uint64
        for {
                a = mid(a)
        }
}
```

```sh
go build -o leafgo leafgo.go
./leafgo & sudo lightswitch --pids $! --duration 5
```

Every sample's real call chain is `leaf ← mid ← main`, but the profile
shows `leaf` directly under `main` with `mid` missing.

---

The PR for this would be more invovled: would mean walking the `pclntab`tables which Go provides and its own runtime uses. Happy to discuss how you want to fix this: one option is to take inspiration from https://github.com/LalitMaganti/sismo/blob/main/crates/sismo-symbolize/src/gopclntab.rs . Though I would *not* advocate for using that directly, sismo as a whole is (intentionally) very vibe-sloppy and everything there will have to be rewritten at some point.

javierhonduco:
I've started implementing a parser for `pclntab` but had to switch gears to something else. I am very interested about this not only for what you mention but also to support [fully stripped Go binaries](https://github.com/javierhonduco/lightswitch/blob/aba1eae172654a589e677f915c9eeb577190a394/src/profiler.rs#L243). I agree with you, it's a pretty foundational part of the code and should be carefully implemented. If you are up for it happy to tag you once I restart the work on the parser?

LalitMaganti:
> If you are up for it happy to tag you once I restart the work on the parser?

Definitely, please do!

PR #531, opened 2026-07-31
Fix Go .text region calculation

javierhonduco:
This was wrong, the mapping addresses were used to calculate the executable address and size but this should be read from the elf data.

Besides causing unwinding to fail this had another worse consequence that I saw on my machine with some workloads written in Go: as the address range differences would be very large, the number pages would be astronomically large, causing a very large memory spike and the BPF map to get full and never be cleaned up since the clean up logic would fail to remove these entries. The only fix would be to restart lightswitch.

This is a semi-temporary fix until proper Go unwinding data is read and used. This is a planned feature, but until then, let's remediate the very large allocation issue here.

Test Plan
=========

CI + before this PR, some custom logging reported this

```
- adding 1533713733 pages for "/nix/store/h9p4xw16i5x5ffdqs7k6h2q93j2ybbhr-moby-29.6.2/libexec/docker/dockerd"
```

### DIFF
--- a/src/profiler.rs
+++ b/src/profiler.rs
@@ -861,6 +861,7 @@ impl Profiler {
         let object_files = self.object_files.read();
         let executable_info = object_files.get(&executable_id).unwrap();
         let executable_path = executable_info.path.clone();
+        let mut elf_loads = executable_info.elf_load_segments.clone();
         let opened_exe_path = if executable_info.is_vdso {
             executable_info.path.clone()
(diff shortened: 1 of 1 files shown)


javierhonduco:
Related to: https://github.com/javierhonduco/lightswitch/issues/524, as implementing what's missing will properly fix both cases

samanpa (approved):

PR #534, opened 2026-08-03
Surface unwind diagnostics in profiles

javierhonduco:
Understanding unwinding errors in profiles is not straightforward. By capturing these events and propagating them for flamegraphs, pprofs, and firefox profiles it can be in great aid in understanding when something doesn't go as expected.

Additionally this has the benefit of resulting in less biased profiles since the samples will show up with the correct weights, even if the stacks are not fully unwound.

Future work
===========

Perfetto doesn't use this field yet, this would be for another PR. Perhaps there's a better way to integrate it with this format like a metadata / event track?

Another future change would be to reuse the statuses for the various counters rather than duplicating the definitions.

Test Plan
=========

CI

### DIFF
--- a/src/profile/sample.rs
+++ b/src/profile/sample.rs
@@ -6,17 +6,67 @@ use anyhow::anyhow;
 use lightswitch_object::ExecutableId;
 use tracing::error;
 
+use crate::bpf::profiler_bindings::sample_result;
 use crate::kernel::KERNEL_PID;
 use crate::process::ObjectFileInfo;
 use crate::process::Pid;
 use crate::process::ProcessInfo;
 use crate::profile::Frame;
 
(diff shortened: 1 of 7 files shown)


src/profile/sample.rs
```diff
@@ -6,17 +6,67 @@ use anyhow::anyhow;
...
+            }
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub struct RawSample {
     pub pid: Pid,
     pub tid: Pid,
+    pub result: SampleResult,
```
samanpa, line 69: Nit add a comment to the struct saying that when a new field is added we have to add to hash

samanpa (approved):

PR #542, opened 2026-08-19
Fix ringbuf size calculation

javierhonduco:
Since the current implementation wasn't totally correct, causing the userspace worker to wake up very often.

Note that there's a potential issue for machines with many CPUs / very high frequency where the program load might be rejected, but I don't anticipate this to be an issue in practice

Test Plan
=========

CI

### DIFF
--- a/src/bpf_objects.rs
+++ b/src/bpf_objects.rs
@@ -163,24 +163,20 @@ impl Bpf {
         inner_map_shape
     }
 
-    fn get_stacks_ringbuf_max_entries(sample_freq: u32) -> u32 {
-        // The assumption here is that although the ringbuf is shared
-        // by all CPUs, it's not expected to get filled up since
-        // 1. At any "single instance", we expect at most n samples to be written
-        // to the ringbuf where n is the number of online cpus emitting events.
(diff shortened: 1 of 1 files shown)


src/bpf_objects.rs
```diff
@@ -163,24 +163,20 @@ impl Bpf {
...
-        let max_entries_bytes: u32 = sample_size_bytes * num_expected_entries;
-
-        // max_entries for ringbuf is required to be specified in bytes, be a multiple
-        // of the page size and a power of two
+    /// Returns the ring buffer size for a given frequency and the CPU count.
+    /// It assumes to never lag behind, since the threads processing the events
+    /// are pretty lightweight.
+    fn ringbuf_size(sample_freq: u32) -> u32 {
+        let num_cpus = get_online_cpus().expect("get online CPUs").len();
+        let num_max_samples = num_cpus.saturating_mul(sample_freq as usize);
```
gmarler, line 171: That's more like it - always worried about the max of online CPUs or frequency being way too  small - as it should be the number of online CPUs * the sampling frequency.

gmarler (approved):
Very nice - and the `saturating_*` calls are just chef's kiss.

PR #548, opened 2026-09-03
Update dependencies

samanpa:


javierhonduco:
On phone right now, didn't review it properly but mind separating the edition bump (and its fixes) and the dependency bump in separate commits?

Happy with the change but curious what motivated the bumps. 

Thank you!

samanpa:
Replaced by #551

PR #551, opened 2026-09-04
Update dependencies

samanpa:
Noticed we were using a version of the lru crate with a known vulnerability https://crates.io/crates/lru/security. So updated all our dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,6 +70,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


javierhonduco (approved):
LGTM! Not sure why some files got reformatted, shouldn't happen if you are using the nix flake? Not a biggie either way, it's not that many files

javierhonduco:
Very strange that I didn't get a dependabot alert, they are enabled...

samanpa:
> Very strange that I didn't get a dependabot alert, they are enabled...

Maybe run `cargo audit`? https://github.com/rustsec/audit-check.

I don't know if dependabot will tell you about new semantic versions

javierhonduco:
It should if they are security releases, I have gotten a bunch of those in the past... Need to look into this

PR #555, opened 2026-09-07
Warn if kASLR reading fails only for `StreamingCollector`

javierhonduco:
As most of the other collectors are used with a local symbolizer.

Test Plan
=========

CI


src/profiler.rs
```diff
@@ -353,17 +354,24 @@ impl Profiler {
...
             }
             Err(e) => {
-                error!(
-                    "fetching the kaslr offset failed with {:?}, assuming it is 0. This ",
+                let msg = format!(
+                    "fetching the kASLR offset failed with `{:?}`, assuming it is 0x0.",
                     e
                 );
+
+                if requires_correct_kaslr {
```
javierhonduco, line 369: Waiting on https://github.com/tokio-rs/tracing/issues/2730

PR #569, opened 2026-09-17
VOID

gmarler:


src/debug_info.rs
```diff
@@ -86,6 +87,50 @@ impl DebugInfoBackendFilesystem {
...
+        }
+    }
+}
+
+/// Parse a `Retry-After` header in delta-seconds form.
+///
+/// The HTTP-date form is not supported: it would need a date parser we do not
+/// need anywhere else. An unparseable value simply means "no hint from the
+/// server".
+fn parse_retry_after(header: Option<&str>) -> Option<u32> {
```
javierhonduco: Does the backend that you use return this header or is this something an LLM wrote "just in case"? I want to be mindful of adding dead code

src/debug_info.rs
```diff
@@ -86,6 +87,50 @@ impl DebugInfoBackendFilesystem {
...
+/// What the backend told us about a build ID.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum BackendQuery {
+    /// The backend has it. Nothing to do.
+    Present,
+    /// The backend definitely does not have it. Upload.
+    Absent,
+    /// We do not know: the backend is shedding load, broken, or unreachable.
+    /// Do *not* upload — a multi-megabyte POST chasing every failed query is
+    /// how a slow backend can turn into a dead one.
```
javierhonduco, line 98: While we should try to have a well-behaved client, ultimately, this should be up to the backend / proxy to shed load

javierhonduco (commented):
Thank you for your PR. I appreciate this effort and the idea behind the PR makes sense, just curious on how was this found. Seems like this was somewhat related to high number "has debug info" type queries we discussed but it's a different case, no?

src/debug_info.rs
gmarler: This enum is in preparation (in a subsequent PR in this 4-PR stream) for stating what the client's behavior/response is when receiving the various HTTP codes from the backend.  The desired behavior on the client side is:
* backend sends 2xx: The client treats the item as Present on the backend (and will, in the next 2 PRs, insert it in the cache as such, not to be tried again during the lifetime of the agent).  Backend does send this.
* backend sends 404: The backend definitely doesn't have it, so we know the debuginfo is Absent on the backend.  Backend sends this today
* backend sends other 4xx - or 5xx unintentionally: In these cases, which the backend sends today, the item of interest is Unavailable - permanently or temporarily, we don't know.  So we'll do a backoff/retry in case the backend is struggling (today) or wants to shed load (in future)

src/debug_info.rs
gmarler: This is something I wrote after seeing another HTTP interchange (which I misremembered as being between the agent and the backend - but was actually from the browser to the profiling UI) - if you don't think the backend will ever send the RETRY_AFTER header back, then we can just default to our own good judgement (MAX_RETRY_AFTER_SECS), sure.

gmarler:
This was found as a result of an empirical measurement of several hosts in clusters that had the highest level of backend debuginfo presence queries over a period of a week.

After applying the complete set of 4 PRs to lightswitch and then running it on that same set of hosts for several days, I've confirmed that:
* The cache (which will appear in PR 3) never exceeds 60 KB
* No debuginfo that's on the backend is ever queried more than once per lightswitch lifetime - this eliminates the vast majority of such queries (> 97%)

src/debug_info.rs
gmarler: And yes, since the backend doesn't currently emit the `Retry-After` header, I'll just remove all references to it.

gmarler:
Unexpected gotcha regarding the PR stacking approach:

I'm contributing from a fork, and GitHub requires a PR's base branch to live in the base repo. Would you be willing to grant push access so the intermediate branches can live here and GitHub can chain them? Happy to scope it however you prefer, and also happy to drop the stacking approach and just submit them one at a time after each merges — it's only a reviewing convenience, not a requirement.

I think I can stack them on my fork just as well, but you won't see them unless you explicitly look for them there.

src/debug_info.rs
```diff
@@ -142,8 +181,12 @@ impl DebugInfoManager for DebugInfoBackendRemote {
 
 impl DebugInfoBackendRemote {
     /// Whether the backend knows about some debug information.
+    ///
+    /// A transport error or a timeout is reported as
+    /// [`BackendQuery::Unavailable`] rather than an error: not being able
+    /// to ask is not conclusive evidence that the backend lacks the build ID.
     #[instrument(level = "debug")]
-    fn find_in_backend(&self, build_id: &BuildId) -> anyhow::Result<bool> {
```
javierhonduco, line 146: Right now this is represented this way:
- HTTP 200: `Ok(true)`
- Any other HTTP: `Ok(false)`
- On error: `Err(_)`

It seems to me that an easier solution would be to check for `StatusCode::NotFound` and return false only in that case and an error otherwise

javierhonduco (commented):
I could not find the other PRs/commits you mentioned on your fork so could not take a look. As mentioned elsewhere I suggest you deploy this in your environment first to ensure it's working as expected.

Two more thoughts:
- adding a cache here can hide code issues, let's be mindful of that;
- the logging you mention would be useful to know what's actually logging;
- this PR could be simplified to a few lines by checking for `NotFound` (see comment)

